PR #4, opened 2020-11-22
Add java time datatypes

bmarinovic:
Issue from `zio-web` project (https://github.com/zio/zio-web/issues/74)
For types that can be constructed without additional info (at least by convention), I left them as `implicit object`. Other types require formatting info so they must be `case class`

### DIFF
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -1,0 +1,37 @@
+version = "2.0.1"
+
+align = most
+assumeStandardLibraryStripMargin = true
+danglingParentheses = true
+docstrings = JavaDoc
+includeCurlyBraceInSelectChains = false
+lineEndings = unix
+maxColumn = 120
+
+continuationIndent {
+  callSite = 2
+  defnSite = 2
+}
+
+newlines {
+  alwaysBeforeTopLevelStatements = true
+  alwaysBeforeMultilineDef = true
+}
+
+spaces {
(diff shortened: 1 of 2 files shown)


jdegoes (approved):

jdegoes:
Looks good, thank you!

These new classes not being implicit has some implications on ease-of-use of Schema. Although I don't see a particular workaround available. For `Duration`, we could actually have wrapper types for every TemporalUnit, and make implicits for those wrapper types available. But the others, not so much.

Now that said, I think ISO8601 could be supported out of the box. 

Perhaps what we can do is to make these implicit but specify a default argument for the date time formatter.

e.g.:

```scala
implicit final case class ZonedDateTime(formatter: DateTimeFormatter = DateTimeFormatter.ISO_LOCAL_DATE_TIME) extends StandardType[ZonedDateTime]
```

This way you would get auto-resolution if you want the common case, or manual resolution if you want a different formatter.

What do you think?

bmarinovic:
Yes, great idea! I was also thinking about giving default argument to case class, but wasn't sure if it would be implicitly resolved :)
Will add that + implicit object `DurationX` for every `TemporalUnit`

jdegoes:
Great! I'd do a little test to make sure Scala can resolve it, something like:

```scala
implicitly[Schema[java.time.LocalDateTime]]
```

If that compiles, we're golden!

PR #8, opened 2020-11-24
Initial commit of wip from https://github.com/zio/zio-web/pull/90. #7

brbrown25:
>
Co-authored-by: Brandon Brown <brandon@bbrownsound.com>
Co-authored-by: Jason Pickens <jasonpickensnz@gmail.com>

Issue #5, opened 2020-11-22
Shift to use ordered maps for Schema.Record and Schema.Enumeration


jdegoes:
Currently, we are using a `Map` for the keys in records and enumerations, which means that we cannot rely on the key ordering. Yet for some protocols, such as grpc, the ordering is very important: it must be stable and ideally reflected by case class / subtype ordering. In order to accommodate this, we need to switch to a map type that allows linear ordering of keys. 

Scala has SortedMap, which could be used, but is not really ideal (it wants to order based on a property of the keys, but that would force us to store `(Int, String)` in the map, which is unpleasant to work with). Possibly we could use an ordinary map, together with an array that maintains key ordering; or we could create our own OrderedMap type.

V-Lamp:
Hey, you could use [ListMap](https://www.scala-lang.org/api/current/scala/collection/immutable/ListMap.html) 
> List map iterators and traversal methods visit key-value pairs in the order they were first inserted.

PR #10, opened 2020-11-29
Add initial ProtobufCodec

eriktim:
Copy from https://github.com/zio/zio-web/pull/98.

This PR adds a Protobuf Codec for encoding & decoding Protobuf byte chunks to/from Schemas (fixes https://github.com/zio/zio-web/issues/75).

* This codec is still using normal `Map`s (see #5).
* `ByteType` is updated to use `Chunk[Byte]` instead of just `Byte`.

I understand that the JSON PR (#8) already contains `Schema` generators. I suggest I update the Protobuf codec tests after those get merged in.

Incomplete parts:

- [x] Add Java time types
- [x] Add decoder failure tests


core/src/main/scala/zio/schema/StandardType.scala
```diff
@@ -15,7 +17,7 @@ object StandardType {
   implicit object LongType   extends StandardType[Long]
   implicit object FloatType  extends StandardType[Float]
   implicit object DoubleType extends StandardType[Double]
-  implicit object ByteType   extends StandardType[Byte]
+  implicit object ByteType   extends StandardType[Chunk[Byte]]
```
jdegoes, line 20: Maybe we should change the name here to `BinaryType` instead of `ByteType`?

core/src/main/scala/zio/schema/codec/ProtobufCodec.scala
```diff
@@ -0,0 +1,485 @@
...
+    ZTransducer(ZManaged.succeed(chunk => ZIO.succeed(Encoder.encodeChunk(schema, chunk))))
+
+  override def decoder[A](schema: Schema[A]): ZTransducer[Any, String, Byte, A] =
+    ZTransducer(ZManaged.succeed(chunk => ZIO.fromEither(Decoder.decodeChunk(schema, chunk))))
+
+  object Protobuf {
+
+    sealed trait WireType {}
+
+    case object VarInt                     extends WireType
```
jdegoes, line 23: I'd put all these in the companion object of `WireType`.

core/src/main/scala/zio/schema/codec/ProtobufCodec.scala
```diff
@@ -0,0 +1,485 @@
...
+      case _: Schema.Primitive[_] => None
+      case _: Schema.Tuple[_, _]  => None
+      case _: Schema.Optional[_]  => None
+    }
+  }
+
+  object Encoder {
+    import Protobuf._
+
+    def encodeChunk[A](schema: Schema[A], chunk: Option[Chunk[A]]): Chunk[Byte] =
```
jdegoes, line 76: What is the `Option[Chunk[A]]`?

core/src/main/scala/zio/schema/StandardType.scala
eriktim: That makes sense. I'll change this.

core/src/main/scala/zio/schema/codec/ProtobufCodec.scala
eriktim: Done.

core/src/main/scala/zio/schema/codec/ProtobufCodec.scala
eriktim: Well, I clearly didn't know how to create a `ZTransducer` properly. So I used a (unneeded) `ZManaged` and the `apply` method.  But having another look now it's the types that kinda force me. I cannot use the `ZTransducer.fromFunction[M]` here as the `Codec` trait tells me to return a `ZTransducer[Any, Nothing, A, Byte]` (encoder) and `ZTransducer[Any, String, Byte, A]` (decoder). Now would it make sense to use `Chunk[Byte]`s here as well? It's not like I can transform a single `Byte` into a value `A`.

eriktim:
@jdegoes can you have another look at this PR? It would be good to have property testing in place as well, but for now I've just added a bunch of tests for all types.

jdegoes (approved):

jdegoes:
@eriktim This looks really awesome. Any changes can be done as followup PRs by myself or others. Thank you for your hard work on this!

eriktim:
Thanks @jdegoes, always happy to contribute!

PR #15, opened 2021-02-27
Support for scala3

scalavision:
This was done in order to learn more about Covariance, it turned out however, that it might be useful for the project itself :-)

Status so far is that,

source code compiles for:

* 2.12.13
* 2.13.4
* 3.0.0-M3

The tests succesfully runs for `2.13.4`, with one test for ZoneId commented out.

There is some implicit problem, has not been able to find out the main cause, ZoneOffset works, which is very similar. One difference is that ZoneId is an abstract java class, might be the reason, but don't know how to solve it.

There are minor compilation errors for the tests in 2.12.13, not sure how to handle those.
There need to be implemented a `macros.scala` for Scala 3 as well, but that should be done in another PR.

Also tried using scala 3 without making things covariant, but it became quickly very complex. This might improve with RC1 though, but it seems covariance helps for typesafety nevertheless.

There are still lots of nasty casts, some of them was added to make scala 3 compile.

The crossbuild setup was inspired by what I learned from the zio build.

If there are other reasons not to use Covariance for the Schema, just close this PR, it was mainly a personal exercise to see how far I could get on this, and I learned a lot :-)


jdegoes:
The essential question of covariance:

If `Dog` is a subtype of `Animal`, then should `Schema[Dog]` be a subtype of `Schema[Animal]`? If yes, then in any place where a `Schema[Animal]` is expected, a `Schema[Dog]` can be provided.

The problem with this: code that has a `Schema[Animal]` is expecting to be able to generate codec for _any_ `Animal`. But if the schema is based on the schema for a `Dog`, it does not know about the structure of, say, a `Cat`, but rather, only the structure of a `Dog`. So the code using a `Schema[Animal]` (widened from `Schema[Dog]`) cannot in fact know the schema of all animals, and therefore, cannot generate codecs for all animals, but rather, only for dogs.

For this reason, we can't make `Schema` covariant: because it will lead to unsafe code. 

Now there is an advance technique I call "splitting invariance" that can be used to make Schema both covariant and contravariant. So you could have things like `Schema[Animal, Dog]` (widened from `Schema[Dog, Dog]`), which could be used to serialize a Dog, and deserialize an Animal; but not to deserialize a Dog, or serialize an Animal. Whether that technique is worth it here, I don't know, but I suspect not.

Hope this makes sense. Closing for now but thanks for your work here, and please take a look at some of the other issues currently open!

scalavision:
Thanks for the explanation, this makes perfectly sense now! Also for mentioning the splitting invariance technique, I tried using to parameters on a second attempt on Sunday, it turned out to be really difficult in practice, hehe yeah.. I think it probably is too little gain in this. This week is really full, but maybe I am able to look into something next week. I am also very interested in the zio-flow project. We'll see :-)

PR #21, opened 2021-03-04
Fixes #20: Add Schema.Fail

### LINKED ISSUE #20: Add Schema.Fail
Schema.Fail will be the schema that always fails. It represents the absence of schema information for a given type.

Something like:

```scala
object Schema {
  final case class Fail[A](message: String) extends Schema[A]
  def fail[A](message: String): Schema[A] = SchemaFail(message)
}
```

thinkharderdev:
Fixes #20


jdegoes (approved):
Nice work!

PR #23, opened 2021-03-05
Fixed #17: Add spec for schema equality

### LINKED ISSUE #17: Define valid equality/hashCode on Schema
Schema must have well-defined equality and hashCode methods.

Note that due to the presence of `Schema#transform`, which stores functions in the ADT, we will have to use reference equality on functions to check for equality and to compute hash code. Reference equality implies equality, but is not necessary for equality, which means that two schemas could in fact be equal, but `equals` will return `false`. That's a known and acceptable limitation of this approach.

We should have a test suite that demonstrates equal schemas compare as equal and hash the same.

thinkharderdev:


thinkharderdev:
Unless I'm missing something I think the default equals/hashCode have us covered.

jdegoes (approved):

Issue #16, opened 2021-03-02
Enhance codec with non-streaming methods


jdegoes:
Currently, `Codec` is defined in a stream-oriented way:

```scala
trait Codec {
  def encoder[A](schema: Schema[A]): ZTransducer[Any, Nothing, A, Byte]
  def decoder[A](schema: Schema[A]): ZTransducer[Any, String, Byte, A]
}
```

This is ideal for many applications, but imposes undue performance penalty on non-streaming code.

To solve this, we should enhance the design of codec as follows:

```scala
trait Codec {
  def encode[A](schema: Schema[A]): A => Chunk[Byte]
  def decode[A](schema: Schema[A]): Chunk[Byte] => Either[String, A]

  def encoder[A](schema: Schema[A]): ZTransducer[Any, Nothing, A, Byte]
  def decoder[A](schema: Schema[A]): ZTransducer[Any, String, Byte, A]
}
```

sviezypan:
I would like to take this one

PR #25, opened 2021-03-06
#16 Enhance codec with non-streaming methods

sviezypan:
@jdegoes 
Current streaming-oriented decoder returns Chunk[A] when run. So when decoding an empty string, it succeeds with Chunk.empty.
This new non-streaming variant would in that case fail with "No bytes to decode" string message. Let me know if something else is desired here.


jdegoes:
@sviezypan Looks great! Please just sign the CLA then this will be good to merge!

PR #28, opened 2021-03-10
Fixes #7: Implement JsonCodec

### LINKED ISSUE #7: Implement JsonCodec using ZIO JSON
Moving this issue here from https://github.com/zio/zio-web/issues/73

thinkharderdev:
Fixes #7 

Still need to work out some of the tests but the codec implementation is done so wanted to get feedback in case I am way off base in the implementation.


thinkharderdev:
I think this is done. Need to get wait for a new version of zio-json to get cut with https://github.com/zio/zio-json/pull/202 included so we can update dependency here. 

Also, creating a real `Gen[Random, ZoneId]` seems to cause catastrophic performance issues any which way I try and do it so I have it defined using `Gen.const(ZoneId.systemDefault())` for now.

thinkharderdev:
Waiting on https://github.com/zio/zio-json/pull/213 to get the JDK8 tests to pass. Also blocked on JDK8 Duration handling due to this bug: https://bugs.java.com/bugdatabase/view_bug.do?bug_id=8054978

jdegoes:
@thinkharderdev I merged your ZIO JSON pull request (thank you!). Can you work around the JDK8 bug?

thinkharderdev:
> @thinkharderdev I merged your ZIO JSON pull request (thank you!). Can you work around the JDK8 bug?

Depends. I can "work around" it by using a Gen that doesn't produce durations which trigger it, but to really fix it we would need to re-implement duration parsing in zio-json.

jdegoes:
@thinkharderdev I have logged the bug [here](https://github.com/zio/zio-json/issues/214) so we can keep track of it and work around it in ZIO JSON, and then possibly, until that's done, we could just avoid generating such Duration by using the `suchThat` / `filter` operator on `Gen`. What do you think?

thinkharderdev:
Yep. Works for me.

jdegoes (approved):

jdegoes:

Issue #18, opened 2021-03-02
Allow pattern matching on Schema for tuples, eithers


jdegoes:
Tuples and eithers (products/sums) are so core to schema that we should bake them in and allow pattern matching against them, rather than encoding them atop Record / Enum.

jdegoes:
A hint getting started:

```scala
final case class SchemaTuple2[A, B](left: Schema[A], right: Schema[B]) extends Schema[(A, B)]
final case class SchemaEither2[A, B](left: Schema[A], right: Schema[B]) extends Schema[Either[A, B]]
```

This way, if you have a `Schema[A]`, you can match against it to see if it's Tuple or Either, and then knowing this, you can extract out left / right sides as appropriate.

sviezypan:
let me take this one

PR #29, opened 2021-03-14
#18 introduce Schema.Either & Schema.Tuple cases

sviezypan:
@jdegoes 
I added Schema.tuple(first, second) and Schema.either(left, right) constructors which encodes schemas to Tuple and EitherSchema data structures, which are then pattern matched when encoding/decoding. 

However I encountered some issues and I am not sure which one of those are intentional - and I am just using this schema wrong - or we need to fix those things. I decided to create a PR to move forward with this and get your opinion on the following:
1. encoding/ decoding of sum type worked only when sum type was sticked inside product type as a param. 
2. Product type with other inner product type fails to encode/decode. 
3. Sum type with inner product type fails to encode/decode.
4. Optional schema with product type or nested optional fails to encode/decode.
5. Either of either of something fails to encode/decode correctly.

I fixed n.1 and for other issues I added a bunch of ignored tests to document them. I expect most of this happens because "nestedFields" function does not handle all schema cases. I can try to fix that - either in this PR or new - but I wanted to check with you first.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,5 +1,2 @@
 * @zio/zio-schema
-.github/workflows/ @softinio @mijicd
-build.sbt @softinio @mijicd
-project/BuildHelper.scala @softinio @mijicd
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,6 +11,7 @@ on:
 jobs:
   lint:
     runs-on: ubuntu-20.04
+    timeout-minutes: 30
     strategy:
       fail-fast: false
     steps:
@@ -27,6 +28,7 @@ jobs:
 
   build:
     runs-on: ubuntu-20.04
+    timeout-minutes: 30
     strategy:
       fail-fast: false
       matrix:
@@ -44,6 +46,7 @@ jobs:
 
   publish:
     runs-on: ubuntu-20.04
+    timeout-minutes: 30
     needs: [build,lint]
     if: github.event_name != 'pull_request'
     steps:
--- a/.github/workflows/mdoc.yml
+++ b/.github/workflows/mdoc.yml
@@ -8,6 +8,7 @@ on:
 jobs:
   publish:
     runs-on: ubuntu-20.04
+    timeout-minutes: 30
     steps:
       - uses: actions/checkout@v2
       - uses: olafurpg/setup-scala@v10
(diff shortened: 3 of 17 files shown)


jdegoes:
Did these limitations exist previously? I don't think you should worry about fixing existing bugs as part of this pull request, although if you'd like to look into those bugs separately, I'd appreciate that a lot.

They do all look like bugs to me.

sviezypan:
@jdegoes sorry for late reply. 
Yes, above listed bugs existed previously, basically ProtobufCodec fails to encode more complicated structures (like nested records). I would like to fix them, but I need a little more time.

Meanwhile I added implementation & tests also for JsonCodec. All looks good there.

sviezypan:
@jdegoes I fixed all the bugs I encountered when writing tests in ProtobufCodecSpec. 
I am listing the main issues fixed inside ProtobufCodec.
By encoding:
- encoded key was added in between all fields inside a record. So Wrapper(TwoInts(150, 151)) failed because it looked like Wrapper(OneInt(150), OneInt(151)) - however thats also not entirely correct, because key for OneInt(151) had a field number 1, so in the end it was decoded to Wrapper(TwoInts(151, 0)) (with 0 being default value) This issue did not come up before because standalone TwoInts(150, 151) has field number None, so no key was encoded.
- sum types did not have any key encoded for them. So it was possible to encode/decode standalone enum, but failed when enum was sticked inside another enum or if it contained case class.
By decoding:
- it was not possible to send just the part of the chunk of bytes to the part of the schema. 
Lets say we have Wrapper(OneInt(150), OneInt(151)) which translates to [10, 3, 8, -106, 1, 18, 3 , 8, -105, 1]. Now 10, 3 are referring to length delimited type with length 3 and field number 1. By looping into schema we get to OneInt(150) schema and we send ALL the remainder of bytes to it. So decoder decodes 8, -106, 1 as 150 and then with 18, 3 it looks for second object field inside OneInt schema - which it cannot find since it has just one int type. So the result of decoded chunk was Wrapper(OneInt(150), OneInt(0)). Solution was to send to inner schema just slice of bytes and collect the result.

So tests are examples of what is working. I would expect more bugs to come up once we start using the protocol buffer codec, but I think for now, this PR is ready to be reviewed.

jdegoes (approved):

jdegoes:
Thank you for your work on this and for fixing some of the existing bugs!

PR #37, opened 2021-04-02
Fixes #2: Case Class Schema

### LINKED ISSUE #2: Specialize caseClass1 - caseClassN
Currently, all the `Schema.caseClass` methods are written in terms of `Record`. This is semantically correct, but not optimally performant, because it involves packing and unpacking into maps.

To maximize performance of codecs generated with `Schema`, we should add `CaseClass` subtypes of `Schema` for various common arities of case classes (1 - 22???). Then as a convenience, we could have a method that converts these case classes to a `Schema.Record`, for cases where performance does not matter that much (e.g. `Schema.CaseClass2#toRecord`.

thinkharderdev:
@jdegoes Can you take a look at this guy and see if this is what you had in mind?

Also, for implementing the Json decoder, can we assume that JSON keys are ordered consistently with the case class declaration. AFAICT this  assumption is made by zio-json but not sure if that is a correct assumption.


thinkharderdev:
JsonCodec is implemented. Still need to do the ProtobufCodec.

core/src/main/scala/zio/schema/Schema.scala
```diff
@@ -27,6 +27,410 @@ object Schema {
...
+    override def toRecord: Record = Record(Map(field1, field2))
+  }
+
+  case class CaseClass3[A1, A2, A3, Z](
+    field1: (String, Schema[A1]),
+    field2: (String, Schema[A2]),
+    field3: (String, Schema[A3]),
+    construct: (A1, A2, A3) => Z,
+    extract: Z => (A1, A2, A3)
+  ) extends CaseClass[Z] {
```
jdegoes, line 52: I think this encoding is almost perfect. After studying it for a while, I can see that `extract` will necessarily allocate a tuple. We can eliminate this overhead by switching to a slightly different encoding:

```scala
 case class CaseClass3[A1, A2, A3, Z](
     field1: (String, Schema[A1]),
     field2: (String, Schema[A2]),
     field3: (String, Schema[A3]),
     construct: (A1, A2, A3) => Z,
     extractField1: Z => A1,
     extractField2: Z => A2,
     extractField3: Z => A3
   ) extends CaseClass[Z] 
```

Then we could pick apart the `Z` into its component pieces without doing any allocations; and we can already, having all the component pieces, construct the `Z` by using `construct` without any allocations.

core/src/main/scala/zio/schema/Schema.scala
```diff
@@ -27,6 +27,410 @@ object Schema {
...
+  case class CaseClass2[A1, A2, Z](
+    field1: (String, Schema[A1]),
+    field2: (String, Schema[A2]),
+    construct: (A1, A2) => Z,
+    extract: Z => (A1, A2)
+  ) extends CaseClass[Z] {
+    override def toRecord: Record = Record(Map(field1, field2))
+  }
+
+  case class CaseClass3[A1, A2, A3, Z](
```
jdegoes, line 46: Here and elsewhere, I'd use:

```scala
final case class
```

Since it does not make sense to override case classes.

core/src/main/scala/zio/schema/Schema.scala
```diff
@@ -27,6 +27,410 @@ object Schema {
...
+    field17: (String, Schema[A17]),
+    field18: (String, Schema[A18]),
+    field19: (String, Schema[A19]),
+    field20: (String, Schema[A20]),
+    field21: (String, Schema[A21]),
+    field22: (String, Schema[A22]),
+    construct: (A1, A2, A3, A4, A5, A6, A7, A8, A9, A10, A11, A12, A13, A14, A15, A16, A17, A18, A19, A20, A21, A22) => Z,
+    extract: Z => (A1, A2, A3, A4, A5, A6, A7, A8, A9, A10, A11, A12, A13, A14, A15, A16, A17, A18, A19, A20, A21, A22)
+  ) extends CaseClass[Z] {
+    override def toRecord: Record =
```
jdegoes, line 429: I love the `toRecord` on all of these, because it gives library authors a convenient way to generically handle all of them, albeit with performance penalty.

core/src/main/scala/zio/schema/Schema.scala
```diff
@@ -27,6 +27,410 @@ object Schema {
   sealed case class Tuple[A, B](left: Schema[A], right: Schema[B])       extends Schema[(A, B)]
   final case class EitherSchema[A, B](left: Schema[A], right: Schema[B]) extends Schema[Either[A, B]]
 
+  sealed trait CaseClass[Z] extends Schema[Z] {
```
jdegoes, line 30: Making sure all the case class types extend this is perfect! This will allow more uniform, type safe handling in user-land code.

core/src/main/scala/zio/schema/codec/ProtobufCodec.scala
```diff
@@ -383,6 +384,7 @@ object ProtobufCodec extends Codec {
         case Schema.Optional(codec)           => optionalDecoder(codec).asInstanceOf[Decoder[A]]
         case Schema.Fail(message)             => (_, _) => Left(message)
         case Schema.EitherSchema(left, right) => eitherDecoder(left, right).asInstanceOf[Decoder[A]]
+        case _                                => ???
```
jdegoes, line 387: Quick fix here and elsewhere will be `case caseClass : CaseClass[A] => decoder(caseClass.toRecord)`

jdegoes:
@thinkharderdev This is great! A few minor comments but I am ready to merge when you're happy with it.

core/src/main/scala/zio/schema/codec/ProtobufCodec.scala
thinkharderdev: This won't work because you end up with a `Decoder[Map[String,_]]` instead of a `Decoder[A]`

I think it shouldn't be too hard to implement the protobuf encoder/decoder for the case class schemas though.

core/src/main/scala/zio/schema/Schema.scala
thinkharderdev: I like this a lot. Should set us up well to implement Optics as well.

thinkharderdev:
@jdegoes Addressed review comments. Protobuf encoder/decoder is still unimplemented though. I can knock that out before we merge or I can do a follow-up PR if you'd rather get this in main

jdegoes (approved):

jdegoes:
@thinkharderdev I will merge now to avoid conflicts (it's a pretty big PR). Thanks for your work on this, and look forward to the follow-up PR!

PR #44, opened 2021-04-17
ProtobufCodec refactoring, sequence decoder fixes

sviezypan:
hi @jdegoes 
This PR mostly contains changes that came up from refactorings that we did during 1on1 session.
- Decoder is now a case class with a run method.
- Decoder does not take WireType as an input, since wiretype can be determined from bytes
- Decoder has a "take" method which can split incoming bytes. As a result whole recordDecoder is a lot cleaner.
- I introduced to Decoder also "loop" method, which helps to decode sequences. It has the following signature
` def loop: Decoder[Chunk[A]] `
The idea is that it would decode all bytes recursively. So for example incoming bytes that contain array of strings would be decoded with Schema[String] and result is collected to Chunk[String]

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,5 +1,2 @@
 * @zio/zio-schema
-.github/workflows/ @softinio @mijicd
-build.sbt @softinio @mijicd
-project/BuildHelper.scala @softinio @mijicd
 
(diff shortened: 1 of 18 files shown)


jdegoes (approved):

jdegoes:
@sviezypan Thanks for your work improving this! 🙏

PR #48, opened 2021-04-26
Add ListSchema for pattern match.

ashprakasan:


core/src/main/scala/zio/schema/Schema.scala
```diff
@@ -26,6 +26,7 @@ object Schema {
   final case class Fail[A](message: String)                              extends Schema[A]
   sealed case class Tuple[A, B](left: Schema[A], right: Schema[B])       extends Schema[(A, B)]
   final case class EitherSchema[A, B](left: Schema[A], right: Schema[B]) extends Schema[Either[A, B]]
+  final case class ListSchema[A](codec: Schema[A])                       extends Schema[List[A]]
```
jdegoes, line 29: We don't have to introduce a new case class here, because you can just `transform` the `Sequence` one. So I'd make the implicit but implement it in terms of the Chunk one.

jdegoes (changes requested):
Let's just implement this in terms of the `Sequence` one.

PR #49, opened 2021-04-26
Fixes #40. Fixes #42. Fixes #43. Fixes 41. Add Schema[Nil.type], Schema[None.type], Schema[ChronoUnit], Schema[Left] and Schema[Right]

### LINKED ISSUE #40: Add Schema[Nil.type]
Nil.type has no information so we can always describe it with a schema. It could be useful to serialize / deserialize empty lists, regardless of their element types. So we can add a (derived) schema for the type.

### LINKED ISSUE #42: Add Schema[None.type]
Since `None.type` is isomorphic to Unit, we should just derive a schema for it from the unit schema.

### LINKED ISSUE #43: Add Schema[java.time.ChronoUnit]
This one can probably be derived and does not need to be added to the standard types.

ashprakasan:


jdegoes (approved):

jdegoes:
🙏

core/src/main/scala/zio/schema/Schema.scala
```diff
@@ -16,24 +19,37 @@ sealed trait Schema[A] { self =>
...
+
+  sealed case class Optional[A](codec: Schema[A]) extends Schema[Option[A]]
+
+  final case class Fail[A](message: String) extends Schema[A]
+
+  sealed case class Tuple[A, B](left: Schema[A], right: Schema[B]) extends Schema[(A, B)]
+
   final case class EitherSchema[A, B](left: Schema[A], right: Schema[B]) extends Schema[Either[A, B]]
 
+  final case class ListSchema[A](codec: Schema[A]) extends Schema[List[A]]
```
jdegoes, line 42: @ashprakasan Sorry I missed this, since it was also in the other PR. We should delete ListSchema and implement the schema for List in terms of the existing Schema.Sequence.

ashprakasan: Just noticed that this [method](https://github.com/zio/zio-schema/blob/main/core/src/main/scala/zio/schema/Schema.scala#L814-L815) already exists.

However in the evaluation of `Remote` (for zio/zio-flow), I need to get `Schema[A]` from `Schema[List[A]]`. Not sure how to do that without introducing ListSchema

PR #53, opened 2021-04-28
Fixes #3: Add Schema.Enum encoding for ADTs

### LINKED ISSUE #3: Specialize enumerations 1 - N
Similar to #2 but for enumerations. Remains to be seen exactly how to do this.

thinkharderdev:
Add Schema.Enum encoding for ADT. Currently only encoding up to Enum3[A1,A2,A3,Z] before falling back to generic EnumN encoding but adding higher arities  is pretty trivial


core/src/main/scala/zio/schema/Schema.scala
```diff
@@ -41,6 +41,13 @@ object Schema {
 
   final case class ListSchema[A](codec: Schema[A]) extends Schema[List[A]]
 
+  final case class Case[A, Z](id: String, codec: Schema[A], construct: A => Z, deconstruct: Z => Option[Z])
```
jdegoes, line 44: I love this, as it should minimize boxing! The type signature is a bit off, however, it should be:

```suggestion
  final case class Case[A, Z](id: String, codec: Schema[A], construct: A => Z, deconstruct: Z => Option[A])
```

(i.e. `Option[A]` in the return value of `deconstruct`).

Second, I wonder if we can do a _bit_ better from a performance standpoint.

What if we require that `A <: Z`? Then, because `A` is already a subtype of `Z`, we get `construct` for free!

That means we don't have to use an identity function or tuple arguments to feed into a case class constructor.

That only leaves the `Option` allocation in `deconstruct`. We could eliminate that, if we wanted, but we would need a low-level, dirty trick that we would have to try to hide:

```scala
final case class Case[A, Z](id: String, codec: Schema[A], construct: A => Z, unsafeDeconstruct: Z => A) {
  def deconstruct(z: Z): Option[A] = try { Some(unsafeDeconstruct(z)) } catch { case _ : IllegalArgumentException => None }
}
```

where we would engineer the low-level `unsafeDeconstruct` to throw an `IllegalArgumentException` if the specified `Z` is in fact not the required subtype `A`.

What do you think?

thinkharderdev: Not sure I fully get the deconstruct change. Is it just so that we can avoid the allocation in the codec decoders? There were can "look up" the case by index or id so we can just call `unsafeDeconstruct` directly. But wouldn't we incur a penalty in the encoders? When we do this:

```scala
    private def enumEncoder[Z](cases: Schema.Case[_, Z]*): JsonEncoder[Z] =
        var matched = false
        cases
          .map(c => c -> c.deconstruct(value))
          .foreach {
            case (c, Some(cast)) =>
              matched = true
              out.write('{')
              val indent_ = bump(indent)
              pad(indent_, out)
              string.unsafeEncode(JsonFieldEncoder.string.unsafeEncodeField(c.id), indent_, out)
              if (indent.isEmpty) out.write(':')
              else out.write(" : ")
              schemaEncoder(c.codec.asInstanceOf[Schema[Any]]).unsafeEncode(cast, indent, out)
              out.write('}')
            case _ => ()
          }
        if (!matched) out.write("{}")
      }
```

for each case that doesn't match we're still allocating the `IllegalArgumentException` and also incurring the exception handling overhead. We could of course use `unsafeDeconstruct` directly but then we are saving the Option allocation on the matching case but also incurring the exception overhead on all the others.

thinkharderdev: Nvm. I see now. We can remove the second allocation by calling unsafeDeconstruct in the Protobuf encoder after we find the case and need to cast before serialization. Added a couple of microbenchmarks and it is ~20-25% higher throughput. 

Made the other changes as well.

benchmarks/src/main/scala/zio/schema/codec/ProtobufBenchmarks.scala
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  @Param(Array("1000"))
+  var size: Int = _
+
+  @Benchmark
+  def enumEncoding(): Chunk[Byte] = {
+    val output = Chunk.empty
+    for (i <- 0 until size) {
+      output ++ statusProtobufEncoder(statuses(i % 3))
```
jdegoes, line 32: This will, to some degree, be measuring the overhead of `++`. Best to take the approach you did in `enumDecoding` and store the individual chunks into an array of chunks.

benchmarks/src/main/scala/zio/schema/codec/ProtobufBenchmarks.scala
```diff
@@ -0,0 +1,46 @@
...
+    }
+    output
+  }
+
+  @Benchmark
+  def enumDecoding(): Array[Status] = {
+    val output = Array.ofDim[Status](size)
```
jdegoes, line 39: Best to push array allocation, etc., into `@setup` so it doesn't affect benchmark (actually this will be quite fast but just in general, a good idea to do setup outside the benchmark proper).

jdegoes (approved):

jdegoes:
@thinkharderdev Excellent work and high attention to performance details. Ready to merge as soon as conflicts fixed!

benchmarks/src/main/scala/zio/schema/codec/ProtobufBenchmarks.scala
thinkharderdev: Good call. Done.

jdegoes (approved):

jdegoes:
Thank you for your work on this! 🙏

PR #55, opened 2021-05-06
Fixes #5: Switch to SortedMap for generic record schemas

### LINKED ISSUE #5: Shift to use ordered maps for Schema.Record and Schema.Enumeration

thinkharderdev:
I think SortedMap gets us all we need here. Since we are now encoding case classes with a dedicated schema which preserves field order we don't have to deal with that use case in the generic Record structure. 

The only thing I had to deal with for case classes is when they have arity 23+ in which case we fall back to the generic map structure. To deal with that I add another package private `Schema.UnorderdRecord` to encode high-arity case classes. The macro derivation should ensure field order across derivations.


jdegoes:
@thinkharderdev I think we should use `ListMap` which will preserve linear ordering, whereas `SortedMap` will sort by the keys, which will result in alphabetical ordering.

Ideally we will have an ordering in the `ListMap` that reflects the declaration order of the fields in the case class, and the declaration order of the cases in the enumeration (sealed trait).

What do you think?

thinkharderdev:
> @thinkharderdev I think we should use `ListMap` which will preserve linear ordering, whereas `SortedMap` will sort by the keys, which will result in alphabetical ordering.
> 
> Ideally we will have an ordering in the `ListMap` that reflects the declaration order of the fields in the case class, and the declaration order of the cases in the enumeration (sealed trait).
> 
> What do you think?

Maybe I misunderstood the problem we are trying to solve with this one. The way I was thinking about it is that we want to preserve standard map semantics with Record schemas (so `Schema.Record(Map("a" -> _, "b" -> _, "c" -> _))` is equivalent to `Schema.Record(Map("c" -> _, "b" -> _, "a" ->_))`). But the protobuf serialization throws away the keys and we need to rely on field order. Using `ListMap` would preserve insertion order but it wouldn't guarantee field order across different instantiations. So if client A builds their Schema.Record in one order and client B in a different order then the protobuf deserialization wouldn't work for both.

jdegoes:
The way I would view it is like this: Schema is a Scala thing and if it exists in a Scala project, the same Schema would be used everywhere, rather than being re-created separately anywhere it is needed. So consistent ordering with respect to a single Schema is important, while consistent ordering across different Schema is not so important.

In addition, in some cases users will be creating case classes from protobuf. The field order is chosen for them. They need the Schema ordering to reflect the ordering in the protobuf.

In still other cases, users may need to add a new field without disrupting the numbers for existing fields, for forward/backward compatibility reasons, which again implies they can choose field ordering.

I think all of these argue for ListMap, even though you are right, it means creating the same Schema with two different case classes with different field orderings would result in two different protobuf codecs.

thinkharderdev:
> The way I would view it is like this: Schema is a Scala thing and if it exists in a Scala project, the same Schema would be used everywhere, rather than being re-created separately anywhere it is needed. So consistent ordering with respect to a single Schema is important, while consistent ordering across different Schema is not so important.
> 
> In addition, in some cases users will be creating case classes from protobuf. The field order is chosen for them. They need the Schema ordering to reflect the ordering in the protobuf.
> 
> In still other cases, users may need to add a new field without disrupting the numbers for existing fields, for forward/backward compatibility reasons, which again implies they can choose field ordering.
> 
> I think all of these argue for ListMap, even though you are right, it means creating the same Schema with two different case classes with different field orderings would result in two different protobuf codecs.

Ah, I think I was confused because I meant to have CaseClassN.toRecord return a `Schema.UnorderedRecord` but didn't actually do that... So the way it is now would scramble the case class field order when converting the case class encoding to a generic record.

What if we did something like this:

```scala
sealed trait Record[A] extends Schema[A] {
    def structure: Map[String,Schema[_]]
}

final case class GenericRecord(override val structure: Map[String,Schema[_]]) extends Record[Map[String,Schema[_]])

final case class CaseClass1[A, Z](field: (String, Schema[A]), construct: A => Z, extractField: Z => A)
      extends Record[Z] {
    override def structure: Map[String,Schema[_]] = ListMap.from(Map(field))
}

  final case class CaseClass2[A1, A2, Z](
    field1: (String, Schema[A1]),
    field2: (String, Schema[A2]),
    construct: (A1, A2) => Z,
    extractField1: Z => A1,
    extractField2: Z => A2
  ) extends Record[Z] {
    override def structure: Map[String, Schema[_]] = ListMap.from(Map(field1, field2))
  }
```
and so on

I think this was the approach you suggested in the CaseClass PR. Guess I should have listened :)

thinkharderdev:
Alright, I think this is good to go now.

jdegoes (approved):
Excellent work. Happy to see this long-standing issue resolved. Thank you! 🙏

PR #59, opened 2021-05-06
Implement Generic

### LINKED ISSUE #36: Add a generic data representation
We should have a generic data representation (`Generic`) that mirrors the `Schema` hierarchy, such that we can convert an `A` and a `Schema[A]` into a `Generic`.

Something like:

```scala
sealed trait Schema[A] {
  def toGeneric(value A): Generic = ???

  def fromGeneric(generic: Generic): Either[String, A] = ???
}
```

where:

```scala
sealed trait Generic
object Generic {
  final case class Integer(value: Int) extends Generic
  final case class Record(values: Map[String, Generic]) extends Generic
  final case class Enumeration(value: (String, Generic)) extends Generic
  ...
}
```

As a law: `schema.fromGeneric(schema.toGeneric(a)) === Right(a)`.

It would be ideal (for #35) if there is an `Ordering` for `Generic`, as well as (of course) consistent equals/hashCode.

adamgfraser:
Work in process to implement `Generic` to fix #36. Currently includes implementation of `Generic` for primitive types and record types. Still need to do implementation for enumerations as well as other more specialized types.

Copying @kitlangton.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,5 @@
+{
+  "files.watcherExclude": {
+    "**/target": true
+  }
+}
--- a/project/BuildHelper.scala
+++ b/project/BuildHelper.scala
@@ -9,7 +9,7 @@ import scalafix.sbt.ScalafixPlugin.autoImport._
 
 object BuildHelper {
 
-  val zioVersion      = "1.0.3"
+  val zioVersion      = "1.0.7"
   val zioNioVersion   = "1.0.0-RC9"
   val zioJsonVersion  = "0.1.2"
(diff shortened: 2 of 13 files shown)


kitlangton:
Should `Enumeration` be `Schema[Map[String, _]]`, or just a Schema[(String, _)]?

```scala
  sealed case class Enumeration(structure: Map[String, Schema[_]]) extends Schema[Map[String, _]]
```

jdegoes:
@kitlangton I think it should probably extend `Schema[(String, _)]` to make it clear there is exactly one of them.

jdegoes:
Superseded by #67

PR #66, opened 2021-05-17
Fixes #45: Capture annotations in record schemas

### LINKED ISSUE #45: Add support for annotations into Schema
Using macros, we should pick up and store annotations on case classes, enumerations (sealed traits), and the individual terms that make up them (fields in case classes, cases of enumerations).

This way, without macros, users of `Schema` can peer into and alter behavior based on annotations.

thinkharderdev:
Fixes #45 
Capture annotations in `Record` schemas:
1. Macro derivation should capture both class and field-level annotations
2. Introduce `Schema.Field[A]` type to encapsulate fields in `Record` schema
3. Encode record fields as `Seq[Schema.Field[_]]` instead of `ListMap[String,Schema[_]]`. By using `ListMap` we lose O(1) lookups anyway so this seemed like the better encoding.


zio-schema/shared/src/main/scala/zio/schema/Schema.scala
```diff
@@ -816,11 +816,14 @@ object Schema {
...
-    def structure: ListMap[String, Schema[_]]
+  final case class Field[A](label: String, schema: Schema[A], annotations: Seq[Any] = Nil)
+
+  sealed trait Record[R] extends Schema[R] {
+    def structure: Seq[Field[_]]
+    def annotations: Seq[Any] = Nil
   }
 
-  final case class GenericRecord(override val structure: ListMap[String, Schema[_]]) extends Record[ListMap[String, _]]
+  final case class GenericRecord(override val structure: Seq[Field[_]]) extends Record[ListMap[String, _]]
```
jdegoes, line 826: It's ok for `Seq` to appear in contravariant position (as an input to a method), but it should not appear in covariant position (as an output from a method) because it provides no _Big Oh_ guarantees on any of its methods.

Instead, I suggest we refactor this to be `Chunk[Field[_]]` here and elsewhere, so that users have guaranteed `O(1)` random access, etc.

zio-schema/shared/src/main/scala/zio/schema/Schema.scala
```diff
@@ -816,11 +816,14 @@ object Schema {
         }
       )
 
-  sealed trait Record[A] extends Schema[A] {
-    def structure: ListMap[String, Schema[_]]
+  final case class Field[A](label: String, schema: Schema[A], annotations: Seq[Any] = Nil)
+
+  sealed trait Record[R] extends Schema[R] {
+    def structure: Seq[Field[_]]
+    def annotations: Seq[Any] = Nil
```
jdegoes, line 823: Ditto for all covariant uses of `Seq`.

zio-schema/shared/src/main/scala/zio/schema/Schema.scala
thinkharderdev: Makes sense

jdegoes (approved):

jdegoes:
@thinkharderdev Thanks for your work on this! One step closer to the Ultimate Goal.

PR #63, opened 2021-05-08
Derive ZIO Test Gen[R, A] from Schema[A]

igrebenik:
Work in progress to derive **zio-test**'s `Gen[R, A]` to fix #56.

@jdegoes 
John could you check if I'm on a right track?

PR #67, opened 2021-05-21
Fixes #36 Complete DynamicValue (previously known as Generic) implementation

### LINKED ISSUE #36: Add a generic data representation

thinkharderdev:
This still may have some issues related to multiply nested Schemas but haven't quite figured out whether the issue is with the way I have implemented the generators or whether it is an implementation issue. Either way I can continue to try and sort it out, but wanted to post this for review. 

This should also make #63 pretty trivial as it is already more or less implemented by allowing derivation of arbitrary `DynamicValues` from a `Schema` (then you can get back to the scala type with `Schema.fromGeneric`).


zio-schema/shared/src/test/scala/zio/schema/SchemaGen.scala
```diff
@@ -301,13 +311,176 @@ object SchemaGen {
...
-    anyPrimitive,
-    anyOptional,
-    anyTuple,
-    anySequence,
-    anyEnumeration,
-    anyRecord,
-    anyTransform,
-    anyEither
+    // anyPrimitive,
+    anyOptional(anyPrimitive)
```
jdegoes, line 315: Dead code or intentional?

zio-schema/shared/src/test/scala/zio/schema/SchemaGen.scala
```diff
@@ -286,9 +296,9 @@ object SchemaGen {
 
   val anyTransformAndValue: Gen[Random with Sized, TransformAndValue[_]] =
     Gen.oneOf[Random with Sized, TransformAndValue[_]](
-      anySequenceTransformAndValue,
-      anyRecordTransformAndValue,
-      anyEnumerationTransformAndValue
+      anySequenceTransformAndValue
+      // anyRecordTransformAndValue,
```
jdegoes, line 300: Dead code or intentional?

zio-schema/shared/src/main/scala/zio/schema/DynamicValue.scala
```diff
@@ -0,0 +1,1066 @@
...
+      case Schema.CaseClass4(_, f1, f2, f3, f4, _, ext1, ext2, ext3, ext4) =>
+        DynamicValue.Record(
+          ListMap(
+            f1.label -> fromSchemaAndValue(f1.schema, ext1(value)),
+            f2.label -> fromSchemaAndValue(f2.schema, ext2(value)),
+            f3.label -> fromSchemaAndValue(f3.schema, ext3(value)),
+            f4.label -> fromSchemaAndValue(f4.schema, ext4(value))
+          )
+        )
+      case Schema.CaseClass5(_, f1, f2, f3, f4, f5, _, ext1, ext2, ext3, ext4, ext5) =>
```
jdegoes, line 213: Since performance is not incredibly important for `DynamicValue`, I think it's probably ok to just base these on the conversion to the generic record case. It can delete a lot of code.

jdegoes:
@thinkharderdev Really nice work! Just a few minor comments, then good to merge.

zio-schema/shared/src/main/scala/zio/schema/DynamicValue.scala
thinkharderdev: yeah, that's what I wanted to do originally but we need the `extractFieldN`s in order to pull values our of the instance of `Z`

zio-schema/shared/src/test/scala/zio/schema/SchemaGen.scala
thinkharderdev: Unintentional. Commented while debugging and forgot to uncomment :( Fixed

zio-schema/shared/src/test/scala/zio/schema/SchemaGen.scala
thinkharderdev: This was intentional. These cause the tests not to halt because of 
```scala
  // TODO: How do we generate a value of a type that we know nothing about?
  val anyEnumerationTransformAndGen: Gen[Random with Sized, EnumerationTransformAndGen[_]] =
    Gen.empty
```

Still need to sort out how to generate arbitraty `Schema.Tranform`s

thinkharderdev: Also, I didn't include in this PR but I think we can actually go back and clean up a lot of the test code now since we should be able to generate arbitrary values really easily and a lot of the existing generators are redundant. 

There is also still the issue of nested `Schema.Optional`. Right now I've kind of hacked around it by not generating `Schema`s that are more the doubly nested and by manually flattening `Option[Option[_]]` when doing equality checks in the tests. But that won't work if you have, for instance, `Right(Some(None))`

jdegoes (approved):

jdegoes:
Thank you! 🙏

PR #69, opened 2021-05-29
POC: Prevent embedding optional schemas

### LINKED ISSUE #68: Rationalize handling of nested Optional schemas
Currently if we have a nested `Optional` schema such as `Schema.Optional(Schema.Optional(Schema.primtive(StandardType.StringType))` we cannot distinguish between the JSONprotobuf encodings of `Some(None)` and `None` which will both be encoded to `null` in JSON (and similarly for protobuf).

As a first order issue this will cause test failures when we do a "round-trip" encoding/decoding of `Some(None` which will encode to `null` and then decode to just plain `None` instead of `Some(None)`.

A couple of options for handling this:

1. Use a type constraint to prevent "nesting" of optional schemas 
```scala
trait Schema[A] {. self =>
  def optional(implicit ev: A <:!< Option[Any]): Schema.Optional[Option[A]]
} 
```
This generalized type constraint doesn't exist in Scala so we'd have to create it

2. Allow nested optional schemas but make it so `Some(None) == None` in the specs.

thinkharderdev:
Fixes #68 

@jdegoes @adamgfraser 
What do you think of this? This will prevent the "embedding" of optional schemas so that `Schema.Optional(Schema.Optional(someSchema))` will not compile. 

There is one drawback (feature?) in that Magnolia will still derive schemas for something like `case class Foo(field: Option[Option[String]])` by falling back to encoding `field` as an `Enum2`. So you end up with a weird JSON encoding of `Foo`. That might actually be good though since the "standard" JSON null encoding is ambiguous anyway (both `Some(None)` and `None` get encoded as `null` in JSON). With this encoding you end up being able to actually differentiate the JSON representations so `Some(None)` becomes `"some": {"none": {}}` whereas `None` becomes `"none":{}`. If there is some semantic distinction between `Some(None)` and `None` then you can actually faithfully preserve the distinction when serializing to JSON.

### DIFF
--- a/zio-schema/shared/src/main/scala/zio/schema/Schema.scala
+++ b/zio-schema/shared/src/main/scala/zio/schema/Schema.scala
@@ -31,7 +31,7 @@ sealed trait Schema[A] {
   /**
    * A symbolic operator for [[optional]].
    */
-  def ? : Schema[Option[A]] = self.optional
+  def ?(implicit ev: CanBeOptional[A]): Schema[Option[A]] = self.optional
 
   /**
    * A symbolic operator for [[zip]].
@@ -46,7 +46,10 @@ sealed trait Schema[A] {
   /**
    * Returns a new schema that modifies the type produced by this schema to be optional.
    */
-  def optional: Schema[Option[A]] = Schema.Optional(self)
+  def optional(implicit ev: CanBeOptional[A]): Schema[Option[A]] = {
+    val _ = ev
+    Schema.Optional(self)
+  }
 
   /**
    * Returns a new schema that combines this schema and the specified schema together, modeling
@@ -144,7 +147,7 @@ object Schema {
   def first[A](codec: Schema[(A, Unit)]): Schema[A] =
     codec.transform[A](_._1, a => (a, ()))
 
-  implicit def option[A](implicit element: Schema[A]): Schema[Option[A]] =
+  implicit def option[A: CanBeOptional](implicit element: Schema[A]): Schema[Option[A]] =
     Optional(element)
 
   implicit def primitive[A](implicit standardType: StandardType[A]): Schema[A] =
@@ -850,7 +853,7 @@ object Schema {
 
   final case class Primitive[A](standardType: StandardType[A]) extends Schema[A]
 
-  final case class Optional[A](codec: Schema[A]) extends Schema[Option[A]]
+  final case class Optional[A: CanBeOptional](codec: Schema[A]) extends Schema[Option[A]]
 
   final case class Fail[A](message: String) extends Schema[A]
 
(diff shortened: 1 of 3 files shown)


zio-schema/shared/src/main/scala/zio/schema/Schema.scala
```diff
@@ -46,7 +46,10 @@ sealed trait Schema[A] {
   /**
    * Returns a new schema that modifies the type produced by this schema to be optional.
    */
-  def optional: Schema[Option[A]] = Schema.Optional(self)
+  def optional(implicit ev: CanBeOptional[A]): Schema[Option[A]] = {
+    val _ = ev
```
thinkharderdev, line 50: There's got to be a better way to do this so the compiler doesn't complain about unused arguments but I don't know how....

jdegoes:
There are two problems:

1. In the presence of polymorphism, the compiler cannot actually prove that a type is not equal to some other type. The best that can be done is prove the compiler does not _know_ if some polymorphic type is equal to another type. So this weakens the guarantee.
2. A lot of code will be bypassed via Schema derivation (as you mention, for Magnolia), and indeed, the existence of `Option[Option[...A]...]` cannot be avoided because it can naturally occur in user-defined data types.

It may be that we can solve this at the codec level; e.g. a Json codec would encode `Option[A]` using `null`, but only if the `A` were itself not `Option`. It can know that up to a limit by introspecting on the schema (assuming the user isn't `Transform`ing to an `Option` from some other type). Then it could adopt a fancier encoding for the nested options.

thinkharderdev:
> There are two problems:
> 
> 1. In the presence of polymorphism, the compiler cannot actually prove that a type is not equal to some other type. The best that can be done is prove the compiler does not _know_ if some polymorphic type is equal to another type. So this weakens the guarantee.
> 2. A lot of code will be bypassed via Schema derivation (as you mention, for Magnolia), and indeed, the existence of `Option[Option[...A]...]` cannot be avoided because it can naturally occur in user-defined data types.
> 
> It may be that we can solve this at the codec level; e.g. a Json codec would encode `Option[A]` using `null`, but only if the `A` were itself not `Option`. It can know that up to a limit by introspecting on the schema (assuming the user isn't `Transform`ing to an `Option` from some other type). Then it could adopt a fancier encoding for the nested options.

On point 2, the way this works actually I think makes the derived schemas work a bit more nicely in cases where a user-defined types has an `Option[Option[X]]` field since Magnolia will fall back to the ADT encoding of Option (i.e. derive a `Schema.Enum2(...)`). So if a user-defined types get a faithful (if a bit unidiomatic) encoding for free.

thinkharderdev:
Want to take this in a different direction

PR #70, opened 2021-06-02
Fixes #65: Add Schema.Lazy to support recursive macro derivation

### LINKED ISSUE #65: Support for recursive types
The current encoding of case classes has a field for each field of the case class, that returns a tuple of `String` and `Schema`. But if the type is recursive or if indirectly it contains classes that have circular dependencies, it will blow up with stack overflow. Example: https://scastie.scala-lang.org/mSCrN0pQRjOw9pBZNzMSAQ

In Caliban I dealt with this by making the field information lazy (`() => FieldInfo` instead of `FieldInfo`).

thinkharderdev:
Fixes #65 Handle recursive macro derivation by wrapping case class fields in a `Schema.Lazy[A](f: () => Schema[A])`


zio-schema/shared/src/main/scala/zio/schema/Schema.scala
```diff
@@ -858,6 +860,8 @@ object Schema {
 
   final case class EitherSchema[A, B](left: Schema[A], right: Schema[B]) extends Schema[Either[A, B]]
 
+  final case class Lazy[A](schema: () => Schema[A]) extends Schema[A]
```
jdegoes, line 863:
```suggestion
  final case class Lazy[A](private val schema0: () => Schema[A]) extends Schema[A] {
    lazy val schema = schema0
  }
```

zio-schema/shared/src/test/scala/zio/schema/DeriveSchemaSpec.scala
```diff
@@ -28,61 +30,82 @@ object DeriveSchemaSpec extends DefaultRunnableSpec {
...
+              (s: Status) => s.asInstanceOf[Pending.type]
+            )
           )
-        )
 
-      assert(derived)(hasSameSchema(expected))
-    }
+        assert(derived)(hasSameSchema(expected))
+      },
+      test("correctly derives for case classes with recursive types") {
```
jdegoes, line 100: So great to see these passing! 🎉 👯‍♂️

jdegoes (changes requested):
Just a minor comment, otherwise looks great!

zio-schema/shared/src/main/scala/zio/schema/Schema.scala
thinkharderdev: Good call!

thinkharderdev: Done!

jdegoes (approved):

jdegoes:
Fantastic work on this. Thank you! 🙏

jdegoes:
@thinkharderdev Looks like docs failed, can you take a look?

<https://github.com/zio/zio-schema/actions/runs/909896564>

thinkharderdev:
> @thinkharderdev Looks like docs failed, can you take a look?
> 
> https://github.com/zio/zio-schema/actions/runs/909896564

Should be fixed with my PR. Never added a website to this repo.

PR #71, opened 2021-06-05
wip: first cut implementation of schema migrate

girishkolantra:
This is work in progress. I would like to get some feedback on whether I am in the right direction here.

### DIFF
--- a/zio-schema/shared/src/main/scala/zio/schema/Schema.scala
+++ b/zio-schema/shared/src/main/scala/zio/schema/Schema.scala
@@ -80,6 +80,13 @@ sealed trait Schema[A] {
    */
   def zip[B](that: Schema[B]): Schema[(A, B)] = Schema.Tuple(self, that)
 
+  def migrate[B](newSchema: Schema[B]): Either[String, A => Either[String, B]] =
+    (self, newSchema) match {
+      case (left: Schema[A], right: Schema[B]) =>
+        Right((a: A) => DynamicValue.fromSchemaAndValue(left, a).toCompatibleTypedValue(right))
+      case _ => Left(s"migration from $self to $newSchema is currently not possible")
+    }
+
 }
 
 object Schema {
--- a/zio-schema/shared/src/main/scala/zio/schema/DynamicValue.scala
+++ b/zio-schema/shared/src/main/scala/zio/schema/DynamicValue.scala
@@ -98,6 +98,101 @@ trait DynamicValue { self =>
         Left(s"Failed to cast $self to schema $schema")
     }
 
+  def toCompatibleTypedValue[A](schema: Schema[A]): Either[String, A] =
+    (self, schema) match {
+      case (DynamicValue.Primitive(value, p), Schema.Primitive(p2)) if p == p2 =>
+        Right(value.asInstanceOf[A])
+
+      case (DynamicValue.Record(values), Schema.GenericRecord(structure)) =>
+        DynamicValue.decodeStructureOfMatchingFields(values, structure).asInstanceOf[Either[String, A]]
+
+      case (DynamicValue.Record(values), s: Schema.Record[A]) =>
+        DynamicValue
+          .decodeStructureOfMatchingFields(values, s.structure)
+          .map(m => Chunk.fromIterable(m.values))
+          .flatMap(s.rawConstruct)
+
+      case (DynamicValue.Singleton(l), Schema.CaseObject(r)) if l == r =>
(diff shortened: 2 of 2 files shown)


zio-schema/shared/src/main/scala/zio/schema/Schema.scala
```diff
@@ -80,6 +80,13 @@ sealed trait Schema[A] {
    */
   def zip[B](that: Schema[B]): Schema[(A, B)] = Schema.Tuple(self, that)
 
+  def migrate[B](newSchema: Schema[B]): Either[String, A => Either[String, B]] =
+    (self, newSchema) match {
+      case (left: Schema[A], right: Schema[B]) =>
+        Right((a: A) => DynamicValue.fromSchemaAndValue(left, a).toCompatibleTypedValue(right))
```
adamgfraser, line 86: @gk I have some concerns about the approach of going through `DynamicValue` here because it causes us to lose information about whether we can do this migration at all.

The return type of `Either[String, A => Either[String, B]]` indicates two failure modes:

1. We could be unable to migrate between these schemas at all. For examples they may be completely unrelated schemas. We are able to determine if this is the case just by looking at the two schemas without examining any actual values described by either of the schemas.
2. We could be able to do the migration in general but be unable to successfully map a particular value in the migration, for example we have two enumerations that are the same except for one containing an extra case so we can migrate in general but can't migrate a value of the extra case to the other schema. This failure occurs only when the migration itself is valid and we examine a particular value of the type described by the schema.

The current approach of going through `DynamicValue` essentially shifts all of the failures to the second category. Notice how you are always returning a `Right` here because the first case of the pattern match is irrefutable. So right now we're basically always saying it is a valid migration and then just failing whenever we can't do it for a particular value, which is throwing away information.

I think we need to go back to the approach of examining the structures of each of the schemas and seeing if they are compatible. There is obviously a risk of a combinatorial explosion of possibilities here so I would suggest the following approach:

1. Only support a very limited number of migrations without a `Hint` that are basically isomorphisms. Maybe initially the only ones that would be supported here would be the identity migration (we can always migrate from a schema to itself), re-ordering fields within a record type, and possibly adding additional cases to an enumeration.
2. Introduce a new data type `Hint` or `Patch` that would basically be a function from schema to schema and would be an algebraic data type with various subtypes such as `RenameField`, `AddField`, `RemoveField`, and so on.
3. Have `migrate` take an additional option parameter for a `Hint` or `Patch`. Then the implementation would apply the changes from the hint or patch and determine whether the patched schema matches the target schema as described in (1) above. If so we succeed, otherwise we fail.

thinkharderdev: Should the migration itself be typed? That is do we want to do something like:

```scala
sealed trait Transformation
case object Identity extends Transformation
case class RenameField[A](field: Field[A], newName: String) extends Transformation
case class RemoveField[A](field: Field[A]) extends Transformation
case class AddField[A](field: Field[A], defaultValue: A0) extends Transformation
case class ComputeField1[A,B](sourceField: Field[A], targetField: Field[B], compute: A => B) extends Transformation
case class ComputeField2[A1,A2,B](sourceField1: Field[A1],sourceField2: Field[A2], targetField: Field[B], compute: (A1,A2) => B) extends Transformation
//et cetera


final case class Migration[A,B](transformations: Chunk[Transformation])

def migrate[B](newSchema: Schema[B]): Either[String,Migration[A,B]] = ...
```

I think at some point it could make sense to be able to have a `Schema[Migration[A,B]]` so the migration itself could be serialized/replayed/recorder/etc.

adamgfraser: I'm not sure the `A` and `B` type parameters are doing any work above since they aren't used anywhere in the signature of `Migration`. Otherwise I think that is generally right though I think the signature of `migrate` should be the other way around:

```scala
def migrate[B](newSchema: Schema[B])(patch: Chunk[Transformation]): Either[String, A => Either[String, B]] =
  ???
```

Given an initial schema and a final schema and a patch that describes how to handle the differences between them we can generate a migration function for concrete instances of the data type.

You could also imagine a `diff` operator with a signature like the one you described above but I don't think that can be computed in general because if we have multiple fields with changed names for example we don't know which old field to map to which new field without external information.

It could also possibly make sense to bake the concept of sequential transformations into `Transformation` itself so you could combine two transformations and get back a new transformation that first applied the left transformation and then the right one.

PR #79, opened 2021-06-14
Fixes #78: Implement diffing between values which have a common schema

### LINKED ISSUE #78: Implemented automatic diff
We should be able to use a `Schema[A]` to automatically diff values of type `A`. 

To do this we should add a method like so
```scala
trait Schema[A] { self =>
  def diff(thisValue: A, thatValue: A): Diff =  DiffAlgorithm.fromSchema(self)(thisValue,thatValue)
}
```

Where we have `DiffAlgorithm` and `Diff` that look something like 

```scala
trait DiffAlgorithm[A] { self =>
  def apply(thisValue: A, thatValue: A): Diff
}

object DiffAlgorithm {
  def fromSchema[A](schema: Schema[A]): DiffAlgorithm[A] = ???
}

sealed trait Diff
object Diff {
  final case object Identical extends Diff
  final case class Number[A : Numeric](distance: A) extends Diff
  final case class Tuple(leftDifference: Diff, rightDifference: Diff) extends Diff
  // etc
}
```

thinkharderdev:
This is still a WIP but the design is (mostly) done. 

This will implement diffing between values of type `A` when a `Schema[A]` is available.

### DIFF
--- a/zio-schema/shared/src/main/scala/zio/schema/Diff.scala
+++ b/zio-schema/shared/src/main/scala/zio/schema/Diff.scala
@@ -1,11 +1,25 @@
 package zio.schema
 
 import java.math.BigInteger
-import java.time.{ DayOfWeek, Instant }
-import java.util.concurrent.TimeUnit
+import java.time.temporal.{ ChronoUnit, Temporal => JTemporal, TemporalAmount, TemporalUnit }
+import java.time.{
+  DayOfWeek,
+  Duration => JDuration,
+  Instant,
+  LocalDate,
+  LocalDateTime,
+  LocalTime,
+  Month => JMonth,
+  MonthDay,
+  OffsetDateTime,
+  OffsetTime,
+  Year,
+  YearMonth,
+  ZoneId,
+  ZonedDateTime
+}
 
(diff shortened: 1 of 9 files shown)


zio-schema/shared/src/main/scala/zio/schema/Diff.scala
```diff
@@ -0,0 +1,209 @@
...
+
+import java.math.BigInteger
+import java.time.{ DayOfWeek, Instant }
+import java.util.concurrent.TimeUnit
+
+import scala.concurrent.duration.TimeUnit
+
+import zio.Chunk
+
+trait DiffAlgorithm[A] { self =>
```
jdegoes, line 11:
```suggestion
trait Differ[A] { self =>
```

jdegoes (commented):
This is really amazing work. Will be useful in so many different ways (e.g. json patching, data ETL, etc.).

thinkharderdev:
Might need a few more unit tests but this is good to go I think now.

jdegoes (approved):
Beautiful work. ❤️

PR #80, opened 2021-06-18
Add defaultValue to Schema 

### LINKED ISSUE #74: Add `defaultValue` to Schema trait
A `Schema#defaultValue: Either[String, A]` method can be added to the trait `Schema`, which computes a "default value" for the specified `Schema`.

e.g.:

```scala
trait Schema[A]
  def defaultValue: Either[String, A]
}
```

This default value can be computed as follows:

1. For records, compute the default value of their fields.
2. For enumerations, pick the first case.
3. For primitives, pick whatever seems natural (e.g. 0 or 0.0 or empty string).

Possibly, we can enhance this later with a `@defaultValue` annotation.

IMax153:
This PR adds a `defaultValue` method to the `Schema` trait, which will return an appropriate value a given `Schema` and closes #74. 

**Disclaimer**: I am incredibly new to both Scala and ZIO, so any and all constructive criticism and feedback is most welcome 🙏. 

## Additional Context

For `StandardType`s, I [defaulted the primitive values](https://github.com/IMax153/zio-schema/blob/25607f8c2d12b70d5a73a399ece9402d477c6193/zio-schema/shared/src/main/scala/zio/schema/StandardType.scala#L10) to the following:

**Basic Primitives**
- `UnitType` -> `()`
- `StringType` -> `""`
- `BoolType` -> `true`
- `ShortType` -> `0`
- `IntType` -> `0`
- `LongType` -> `0`
- `FloatType` -> `0.0`
- `DoubleType` -> `0.0`
- `BinaryType` -> `Chunk.empty`
- `CharType` -> `'\u0000'` (the empty character `''` does not compile, so I went with the `NUL` character code)
- `BigDecimalType` -> `BigDecimal.ZERO`
- `BigIntegerType` -> `BigInteger.ZERO`

**`java.time` Primitives**
- `DayOfWeekType` -> `java.time.temporal.WeekFields.of(java.util.Locale.getDefault).getFirstDayOfWeek`
- `Month` -> `java.time.Month.JANUARY`
- `MonthDay` -> `java.time.MonthDay.of(java.time.Month.JANUARY, 1)`
- `Period` -> `java.time.Period.ZERO`
- `Year` -> `java.time.Year.now`
- `YearMonth` -> `java.time.YearMonth.now`
- `ZoneId` -> `java.time.ZoneId.systemDefault`
- `ZoneOffset` -> `java.time.ZoneOffset.UTC`
- `Duration` -> `java.time.Duration.ZERO`
- `Instant` -> `java.time.Instant.EPOCH`
- `LocalDate` -> `java.time.LocalDate.now`
- `LocalTime` -> `java.time.LocalTime.MIDNIGHT`
- `LocalDateTime` -> `java.time.LocalDateTime.now`
- `OffsetTime` -> `java.time.OffsetTime.now`
- `OffsetDateTime` -> `java.time.OffsetDateTime.now`
- `ZonedDateTime` -> `java.time.ZonedDateTime.now`

I also added a [new test suite](https://github.com/IMax153/zio-schema/blob/feature/default-value/zio-schema/shared/src/test/scala/zio/schema/DefaultValueSpec.scala) (`DefaultValueSpec`) to the current array of test suites.


zio-schema/shared/src/main/scala/zio/schema/Schema.scala
```diff
@@ -828,9 +832,11 @@ object Schema {
     override def toString: String = s"Field($label,$schema)"
   }
 
-  sealed trait Record[R] extends Schema[R] {
+  sealed trait Record[R] extends Schema[R] { self =>
     def structure: Chunk[Field[_]]
     def annotations: Chunk[Any] = Chunk.empty
+    // TODO: make this much more robust
+    def defaultValue: Either[String, R] = self.rawConstruct(structure.map(_.schema.defaultValue.getOrElse(())))
```
IMax153, line 839: I am not sure what the best method is for lifting a `Chunk[Either[String, A]]` to an `Either[String, Chunk[A]]`.

In `fp-ts`, I would generally use `Traverse` to take care of this, but wasn't sure what the most semantic/idiomatic way to do this in ZIO was. 

I also [asked the question in Discord](https://discord.com/channels/629491597070827530/629491597070827534/854746145191231538) and was informed about `zio-prelude`'s analogous methods, but I wanted to avoid adding additional dependencies to `zio-schema`.

zio-schema/shared/src/main/scala/zio/schema/Schema.scala
```diff
@@ -843,50 +849,102 @@ object Schema {
...
+  }
 
-  sealed case class Enumeration(structure: ListMap[String, Schema[_]]) extends Schema[(String, _)]
+  sealed case class Enumeration(structure: ListMap[String, Schema[_]]) extends Schema[(String, _)] {
+
+    def defaultValue: Either[String, (String, _)] =
+      if (structure.isEmpty)
+        Left(s"cannot access default value for enumeration with no members")
+      else
+        structure.head._2.defaultValue.map((structure.head._1, _))
```
thinkharderdev, line 862: I think in this case we should require a default case be specified when the `Enumeration` is constructed. So maybe change to
```scala
final case class Enumeration(structure: ListMap[String,Schema[_]], defaultCase: Option[String] = None) extends Schema[(String,_)] {
   def defaultValue: Either[String,(String,_)] = 
      defaultCase match {
         case Some(id) => structure.get(id).map(_.defaultValue.map(id -> _)).getOrElse(Left("invalid default case"))
         case None => Left("no default case is defined")
      }

File: zio-schema/shared/src/main/scala/zio/schema/Schema.scala [hunk 3]
```diff
@@ -843,50 +849,102 @@ object Schema {
...
+    def defaultValue: Either[String, Z] =
+      if (cases.isEmpty)
+        Left("cannot access default value for enum with no members")
+      else
+        cases.head.codec.defaultValue.asInstanceOf[Either[String, Z]]
+  }
+
+  final case class CaseObject[Z](instance: Z) extends Schema[Z] {
+    def defaultValue: Either[String, Z] = Right(instance)
+  }
```
thinkharderdev, line 937: Same here. It feels a little wonky to just declare on subtype as the default but these are also constructed using macros so we can't really allow one to be declared by the user (we could, but it would require an annotation which hasn't been defined yet). Maybe for these we just have `defaultValue` return `Left("no default value can be determined"). If we want to allow a default to be specified in the future then we can do it with an annotation.

zio-schema/shared/src/main/scala/zio/schema/StandardType.scala
```diff
@@ -1,43 +1,130 @@
...
+  implicit object UnitType extends StandardType[Unit] {
+    def defaultValue: Either[String, Unit] = Right(())
+  }
+
+  implicit object StringType extends StandardType[String] {
+    def defaultValue: Either[String, String] = Right("")
+  }
+
+  implicit object BoolType extends StandardType[Boolean] {
+    def defaultValue: Either[String, Boolean] = Right(true)
```
thinkharderdev, line 23: I think Bool should default to `false` to match the behavior of an uninitialized `Boolean` in Java and Scala.

thinkharderdev (changes requested):
A few comments but overall nicely done! Thanks for your work on this!

thinkharderdev:
@jdegoes Is there any particular reason we want `defaultValue` to return `Either[String,A]`. My instinct would be to have it return `Option[A]` (i.e. there either is a default value or there isn't but there shouldn't really be any way an error can occur retrieving it)?

zio-schema/shared/src/main/scala/zio/schema/Schema.scala
jdegoes: I would suggest we just use the **first** subtype. Or we could do the "smallest" subtype, but first is more predictable.

Both Haskell and Scala type class derivation libraries choose to order the terms in a sum type (for `Ord[A]`) according to declaration order, so there's a precedent for this.

zio-schema/shared/src/main/scala/zio/schema/StandardType.scala
jdegoes: 👍

jdegoes:
@thinkharderdev The only reason it has to return `Either[String, A]` is because of `Transform` nodes, which may fail for arbitrary reasons. In which case it's nice to preserve the error.

zio-schema/shared/src/main/scala/zio/schema/Schema.scala
jdegoes: I'd suggest avoiding to add ZIO Prelude just yet since it doesn't yet have a 1.0 release. This implementation looks good.

IMax153:
Thank you @thinkharderdev and @jdegoes for your review and suggestions! Apologies for the delay - I have been on vacation.

I made the requested change to the `BoolType` default value (now defaults to `false`), and used the implementation suggested by @thinkharderdev for lifting a `Chunk[Either[String, A]]` to an `Either[String, Chunk[A]]`.

I also left the `defaultValue` implementations for both `Enumeration` and `Enum` to default to the first declared subtype, but am happy to adjust if necessary.

Please let me know if you think any additional changes are necessary!

Also as an aside, I am happy to work on a `@defaultValue` annotation as well either as part of this PR or as a separate PR - if you have any good resources on how to create an annotation, I can get started on this.

jdegoes:
@IMax153 Thank you for your work on this! Can you resolve conflicts so we can see if the automated build passes?

IMax153:
@jdegoes - thank you! I believe I've resolved all conflicts.

Also, as I mentioned - if you guys would like me to try to add a `@defaultValue` annotation as part of this PR, I can try to work on it - would probably just need an example from the ZIO ecosystem to get me started.

thinkharderdev (approved):

PR #83, opened 2021-06-27
Fixes #12: Implement Schema.serializable

thinkharderdev:
Create a `MetaSchema` data structure to represent the abstract structure of a schema and an associated `Schema.Meta(spec: MetaSchema) extends Schema[Schema[_]]` which is the "serialized" representation of a `Schema[A]`. 

This is about 90% implemented. The only outstanding items are to support diffing of `Schema.Meta` and add some additional unit tests.

Limitations and known issues:
1. The main limitation is that we lose all information about a `Schema[A]` except for the structure when serializing it (since we cannot serialized transform, extractFieldN, deconstruct, construct lambdas). As such the reified `Schema` that you get when you deserialize a `Schema.Meta` is always going to be a generic version of the original (i.e. `Schema.CaseClassN` -> `Schema.GenericRecord` and `Schema.EnumN` -> `Schema.Enumeration`).
2. It's not really obvious what we should do with `Schema.CaseObject` and the current approach discards all type information about the original `object` represented by the schema.


jdegoes:
@thinkharderdev Thanks for working on this!

1. Rather than add a new subtype of Schema, I think you can implement `Schema#serializable` just by dynamically building up a new `Schema` which does not include `Transform` nodes (and which genericizes enums and records, I suppose). Now, this is not "type safe", because there is no way to know, looking at the schema, that it can be serialized and deserialized. But if we want to address that, we can in other ways (e.g. phantom type, path-dependent type, fixed point data, etc.).
2. I think the problem with case object is suggesting we do not have a correct representation for that. Probably, we need to modify `CaseObject` to store an id (as a String?), and then transform it. Or maybe change it to mean some fixed, static value, which can then be transformed to the case object. The key thing is we are missing a generic notion of "this is fixed, static data" in a way that doesn't bundle it with a user-defined data type; whereas for records and enumerations, we have the ability to "lose" the user-defined data types by genericizing (converting to generic records / generic enumerations with just the labels and schemas for the terms).

jdegoes:
Related:

<https://github.com/zio/zio-json/pull/307>

Case objects are isomorphic to unit. So maybe that's another way to handle them: represent them as unit, which is then transformed to the case object. In a sum of case objects, the term ids would be the case object names (names of the subtype of the sum type).

zio-schema/shared/src/main/scala/zio/schema/MetaSchema.scala
```diff
@@ -0,0 +1,128 @@
...
+  def fromSchema[A](schema: Schema[A]): MetaSchema = schema match {
+    case Schema.Primitive(StandardType.Duration(units)) => Duration(units)
+    case Schema.Primitive(tpe)                          => Value(tpe)
+    case Schema.Optional(schema)                        => Optional(fromSchema(schema))
+    case Schema.EitherSchema(left, right)               => EitherMeta(fromSchema(left), fromSchema(right))
+    case Schema.Tuple(left, right)                      => TupleMeta(fromSchema(left), fromSchema(right))
+    case Schema.Sequence(schema, _, _)                  => Sequence(fromSchema(schema))
+    case Schema.Fail(message)                           => Fail(message)
+    case Schema.Transform(schema, _, _)                 => fromSchema(schema)
+    case lzy @ Schema.Lazy(_)                           => fromSchema(lzy.schema)
```
jdegoes, line 37: We should test this on a schema like JSON which is recursive. I think our "schema language" may have to embed references in order to handle recursion (or some sort of "fixed point" operator).

zio-schema/shared/src/main/scala/zio/schema/MetaSchema.scala
```diff
@@ -0,0 +1,128 @@
...
+  object Value {
+    implicit val schema: Schema[Value] = Schema[String].transformOrFail(
+      termType => StandardType.fromString(termType).map(Value(_)).toRight(s"$termType is not valid type"),
+      value => Right(value.valueType.tag)
+    )
+  }
+  final case class Duration(units: TemporalUnit) extends MetaSchema
+
+  object Duration {
+    implicit val schema: Schema[Duration] = Schema[String].transformOrFail(
```
jdegoes, line 113: Really nice!

zio-schema/shared/src/main/scala/zio/schema/MetaSchema.scala
```diff
@@ -0,0 +1,128 @@
...
+        field1 = Schema.Field("id", Schema[String]),
+        field2 = Schema.Field("subType", Schema.defer(MetaSchema.schema)),
+        construct = (label: String, subType: MetaSchema) => MetaCase(label, subType),
+        extractField1 = _.id,
+        extractField2 = _.subType
+      )
+    implicit val chunkSchema: Schema[Chunk[MetaCase]] = Schema.chunk(schema)
+  }
+
+  final case class Value(valueType: StandardType[_]) extends MetaSchema
```
jdegoes, line 102: Love the added type safety. ❤️

zio-schema/shared/src/main/scala/zio/schema/MetaSchema.scala
thinkharderdev: Yeah, that is going to be a problem. I can work on this today.

jdegoes: The [JSON Schema](https://json-schema.org/specification.html) doc uses "pointers" for recursion.

To track recursion, you can use a map based on object identity. This will allow you to detect if you processed the same field before. It might also be possible to define equals / hashCode on the ordinary `Schema` classes in such a way that it terminates even for recursive structures.

thinkharderdev:
Went back to the drawing board and came up with what I think is a much better encoding for the schema serialization. 

Modelling the the "meta schema" as an AST now which has two nice benefits:
1. We can handle recursive references when we are building the tree and replace recursive subtrees with a pointer to the ancestor. 
2. The resulting JSON encoding is more intuitive and human-readable. I think it will more naturally transform to other encodings (JSON schema, Avro schema, etc) as well (I think....)

Also fixed the issue with JSON serialization/deserialization which turns out was just a regular bug. Still having an issue with the protobuf codecs on recursive data types but I assume that is just a bug as well. Need to do some more debugging though to be sure.

zio-schema/shared/src/test/scala/zio/schema/SchemaAssertions.scala
```diff
@@ -22,40 +22,39 @@ object SchemaAssertions {
...
     case (expected, Schema.Transform(actualSchema, _, _)) =>
-      equalsAst(expected, actualSchema)
+      equalsAst(expected, actualSchema, depth)
     case (Schema.Transform(expected, _, _), actual) =>
-      equalsAst(expected, actual)
-    // TODO We use lazy schemas in part to avoid infinite recursion so it's not clear how we test equality without causing StackOverflows
-    case (_: Schema.Lazy[_], _) => true // equalsAst(expected.schema, actual)
-    case (_, _: Schema.Lazy[_]) => true // equalsAst(expected, actual.schema)
-    case _                      => false
+      equalsAst(expected, actual, depth)
```
jdegoes, line 54: Laziness can be tested via reference equality, assuming we are not making mistakes anywhere. It should not be tested for structural equality since that will lead to infinite checks (= stack overflows).

thinkharderdev: I think reference equality isn't what we want here. We need to test that that `expected` and `actual` have the same AST. They may not have (and in most cases we care about w/r/t tests will not) have the same type.

zio-schema/shared/src/main/scala/zio/schema/Ast.scala
```diff
@@ -0,0 +1,293 @@
+package zio.schema
+
+import scala.collection.immutable.ListMap
+
+import zio.{ Chunk, ChunkBuilder }
+
+sealed trait Ast { self =>
```
jdegoes, line 7: My main concern here is with the name `Ast`. Because we are in the `zio.schema` package, it would be nice if we can shoot for `import zioi.schema._` in most code bases. But the name `Ast` is quite common.

zio-schema/shared/src/main/scala/zio/schema/Ast.scala
```diff
@@ -0,0 +1,293 @@
...
+      s.structure
+        .foldLeft(NodeBuilder(s.hashCode(), Chunk.empty)) {
+          case (node, (id, schema)) =>
+            node.addLabelledSubtree(id, schema)
+        }
+        .buildSum()
+    case Schema.Meta(ast) => ast
+  }
+
+  def subtree(schema: Schema[_], lineage: Chunk[Int], optional: Boolean = false, dimensions: Int = 0): Ast =
```
jdegoes, line 129: Can we make this one package private?

zio-schema/shared/src/main/scala/zio/schema/Ast.scala
```diff
@@ -0,0 +1,293 @@
...
+                case (node, (id, schema)) =>
+                  node.addLabelledSubtree(id, schema)
+              }
+              .buildSum()
+          case Schema.Fail(message) => FailNode(message)
+          case Schema.Meta(ast)     => ast
+        }
+      }
+
+  def materialize(ast: Ast, refs: Map[Int, Ast] = Map.empty): Schema[_] = {
```
jdegoes, line 171: Package private?

zio-schema/shared/src/main/scala/zio/schema/Ast.scala
```diff
@@ -0,0 +1,293 @@
...
+      case (true, n) =>
+        (0 until n).foldRight[Schema[_]](baseSchema)((_, schema) => schema.repeated).optional
+    }
+  }
+
+  implicit lazy val schema: Schema[Ast] = DeriveSchema.gen[Ast]
+
+}
+
+object AstRenderer {
```
jdegoes, line 212: Package private?

zio-schema/shared/src/main/scala/zio/schema/Ast.scala
```diff
@@ -0,0 +1,293 @@
...
+      case (false, 0) => baseSchema
+      case (true, 0)  => baseSchema.optional
+      case (false, n) =>
+        (0 until n).foldRight[Schema[_]](baseSchema)((_, schema) => schema.repeated)
+      case (true, n) =>
+        (0 until n).foldRight[Schema[_]](baseSchema)((_, schema) => schema.repeated).optional
+    }
+  }
+
+  implicit lazy val schema: Schema[Ast] = DeriveSchema.gen[Ast]
```
jdegoes, line 208: Maybe expand for long-term stability? What do you think?

zio-schema/shared/src/main/scala/zio/schema/Ast.scala
thinkharderdev: Yeah, that's a good paint. Maybe `SchemaAst`?

jdegoes (approved):

jdegoes:
A giant undertaking. Let's do any revisions as followups. Thanks for your work on this!

PR #91, opened 2021-07-18
Add primitive uuid support

paulpdaniels:
Adds a new primitive for UUIDs. While there is no special support in protobuf or json for them, other potential codec's like BSON have special handling for UUID's.

### DIFF
--- a/zio-schema/shared/src/main/scala/zio/schema/codec/ProtobufCodec.scala
+++ b/zio-schema/shared/src/main/scala/zio/schema/codec/ProtobufCodec.scala
@@ -12,6 +12,7 @@ import zio.stream.ZTransducer
 import zio.{ Chunk, ZIO }
 
 import java.util.UUID
+import scala.util.control.NonFatal
 
 object ProtobufCodec extends Codec {
   override def encoder[A](schema: Schema[A]): ZTransducer[Any, Nothing, A, Byte] =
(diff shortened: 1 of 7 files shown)


zio-schema/shared/src/main/scala/zio/schema/codec/ProtobufCodec.scala
```diff
@@ -587,6 +590,7 @@ object ProtobufCodec extends Codec {
         case StandardType.DoubleType => doubleDecoder
         case StandardType.BinaryType => binaryDecoder
         case StandardType.CharType   => stringDecoder.map(_.charAt(0))
+        case StandardType.UUIDType   => stringDecoder.map(UUID.fromString)
```
thinkharderdev, line 593: Here I think we need to handle an invalid UUID string and return a `Left("invalid UUID")` So something like 

```scala
stringDecoder.flatMap { uuidString =>
  try {
   succeed(UUID.fromString(uuidString))
  } catch { case _: Throwable => fail("invalid UUID") }
}
```

thinkharderdev (changes requested):
Thanks for you work on this! One small comment.

zio-schema/shared/src/main/scala/zio/schema/codec/ProtobufCodec.scala
paulpdaniels: Ok will do, wasn't sure because I noticed that there wasn't any error handling on the `Date` types so I wasn't sure if that was supposed to be handled higher up.

paulpdaniels: Updated! I had to add a new `succeedNow` to avoid the by-name parameter passing which would bypass the try catch.

thinkharderdev: Looks great! I think you just need to run `sbt prepare` to make sure everything is formatted and we should be good to go.

paulpdaniels:
Sorry this fell off my radar for a bit, I'll update this and get it ready for a second pass this weekend

zio-schema/shared/src/main/scala/zio/schema/codec/ProtobufCodec.scala
paulpdaniels: Done!

thinkharderdev (approved):
Thanks for your work on this!

PR #92, opened 2021-07-19
84 Implement Enum4 - Enum22

jupposessho:


zio-schema/shared/src/main/scala/zio/schema/DynamicValue.scala
```diff
@@ -44,6 +45,139 @@ trait DynamicValue { self =>
...
+          case Some(c) => value.toTypedValue(c.codec)
+          case None    => Left(s"Failed to find case $key in enum $s")
+        }
+
+      case (DynamicValue.Enumeration((key, value)), s @ Schema.Enum5(case1, case2, case3, case4, case5)) =>
+        Chunk(case1, case2, case3, case4, case5)
+          .find(_.id == key) match {
+          case Some(c) => value.toTypedValue(c.codec)
+          case None    => Left(s"Failed to find case $key in enum $s")
+        }
```
thinkharderdev, line 60: All of the `EnumX` case classes should extend `Enum` now so you should be able to handle all of these with a single case:
```scala
case (DynamicValue((key,value)), s: Schema.Enum[A]) => 
   s.structure.get(key) match {
      ....
   }
```

thinkharderdev (changes requested):
Thanks! In general when we are not concerned about optimizing performance (i.e. outside of codecs and in test code) we can use the "generic" encoding `Schema.Enum` so we don't have to handle every specialized `Schema.EnumX` case separately.

zio-schema/shared/src/main/scala/zio/schema/DynamicValue.scala
jupposessho: Removed the unnecessary enums, however I had to keep them in `fromSchemaAndValue` and `anyDynamicValueOfSchema` to make the compiler happy.

thinkharderdev (approved):
Thanks!

PR #116, opened 2021-08-09
Type safety improvement for generic records

thinkharderdev:
Create a more type safe encoding for `Schema.GenericRecord` types. The gist is that we preserve field-level type Info with two `HList`-like structures:

`FieldSet` which encodes the record fields and `Struct` which encodes the values in an `HList` like thing. So `GenericRecord(structure: Chunk[Field[_]]) extends Schema[ListMap[String,_]]` becomes `GenericRecord(fieldSet: FieldSet) extends Schema[Struct]` 

This is just a POC now to see if I could get the Json codec working with this encoding (which I did!) so `GenericRecord` is still constructed with a `Chunk[Field[_]]` and we build the `FieldSet` internally but if this works then I think we do away with the `Chunk[Field[_]]` entirely. 

This will require quite a bit more spit and polish before it's done, but wanted to see if this made sense at all before investing more time in it.

### DIFF
--- a/.github/workflows/mdoc.yml
+++ b/.github/workflows/mdoc.yml
@@ -11,7 +11,7 @@ jobs:
     timeout-minutes: 30
     steps:
       - uses: actions/checkout@v2
-      - uses: olafurpg/setup-scala@v10
+      - uses: olafurpg/setup-scala@v12
       - run: sbt docs/docusaurusPublishGhpages
         env:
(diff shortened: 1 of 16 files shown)


thinkharderdev:
Also, eventually we would also replace the generic `Enumeration` encoding with something like `CaseSet` and `Union`.

PR #117, opened 2021-08-10
[WIP] 35 - schema to ordering

eHig:
Here's an initial attempt. Let me know if I'm on the right track.

Period is not comparable. What is the preferred solution here: always return 0; compare assuming 30-day months; or compare years then months then days?

### DIFF
--- a/zio-schema/shared/src/main/scala/zio/schema/SchemaOrdering.scala
+++ b/zio-schema/shared/src/main/scala/zio/schema/SchemaOrdering.scala
@@ -38,13 +38,13 @@ object SchemaOrdering {
     case (Schema.Fail(_), Error(lVal), Error(rVal) ) => lVal.compareTo(rVal)
     case (Schema.Transform(schemaA,_,_), Transform(lVal), Transform(rVal) ) =>
       compareBySchema(schemaA)(lVal,rVal)
-    case (Schema.Enum1(c),_,_) => compareBySchema(Schema.EnumN(Seq(c)))(l,r)
-    case (Schema.Enum2(c1,c2),_, _) => compareBySchema(Schema.EnumN(Seq(c1,c2)))(l,r)
-    case (Schema.Enum3(c1,c2,c3),_, _) => compareBySchema(Schema.EnumN(Seq(c1,c2,c3)))(l,r)
-    case (Schema.EnumN(cases), Enumeration((lField,lVal)), Enumeration((rField,rVal))) if lField==rField =>
-        compareBySchema(cases.find(_.id==lField).get.codec)(lVal,rVal)
-    case (Schema.EnumN(cases), Enumeration((lField,_)), Enumeration((rField,_))) =>
-        cases.indexWhere(_.id ==lField).compareTo(cases.indexWhere(_.id ==rField))
+    case (e:Schema.Enum[_], Enumeration((lField,lVal)), Enumeration((rField,rVal)))  if lField==rField => {
+      compareBySchema(e.structure(lField))(lVal,rVal)
+    }
+    case (e:Schema.Enum[_], Enumeration((lField,_)), Enumeration((rField,_))) => {
+      val fields = e.structure.keys.toList
+      fields.indexOf(lField).compareTo(fields.indexOf(rField))
+    }
     case (r:Schema.Record[_],Record(lVals),Record(rVals)) =>
       compareRecords(r, lVals, rVals)
     case _ => 0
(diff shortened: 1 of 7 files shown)


zio-schema/shared/src/main/scala/zio/schema/StandardType.scala
```diff
@@ -82,51 +85,158 @@ object StandardType {
...
+    override def tag = Tags.FLOAT
+  }
+
+  implicit object DoubleType extends StandardType[Double]      {
+    override def compare(x: Double, y: Double): Int = x.compareTo(y)
+    override def tag = Tags.DOUBLE
+  }
+
+  implicit object BinaryType extends StandardType[Chunk[Byte]] {
+    override def compare(x: Chunk[Byte], y: Chunk[Byte]): Int = {
```
thinkharderdev, line 129: Here I think you can just `x.sum.compare(y.sum)`

thinkharderdev (commented):
Thanks for your work on this! 

I think you are generally on the right track. To give some thoughts on your questions:

I definitely wouldn't return 0 always when comparing `Period`. Comparing years then months then days seems wrong too if you do it lexicographically. My initial instinct is to either
1. Choose a reference date (say the epoch) and compare Periods in reference to that date. 
2. Punt and require the user to supply an implicit `Ordering[Period]` 

2 is probably more correct but would probably impact usability @jdegoes what do you think?

For testing I think you are on the right track as well. I would probably try and create a `Gen` that generates ordered pairs. So something like:

```scala
type PrimitiveAndOrderedPair[A] = (Schema.Primitive[A], Gen[Random with Sized, (A,A)]

val orderedPrimitiveGen: Gen[Random with Sized, PrimitiveAndOrderedPair[_]] = 
  for {
     (schema,gen) <- anyPrimitiveAndGen
     pair <- (gen <*> gen).filter(p => p._1.compare(p._2) < 0
    } yield schema -> pair
```

Then we could recursively build larger structures like we do in `Schema.anyTree` to deterministically test the ordering laws.

zio-schema/shared/src/main/scala/zio/schema/StandardType.scala
eHig: updated this

eHig:
Put in the fixed-start comparison for period for now.
I changed the primitive test, and did a few samples of structures by generating ordered pairs. Is that what you had in mind?

thinkharderdev:
> Put in the fixed-start comparison for period for now.
> I changed the primitive test, and did a few samples of structures by generating ordered pairs. Is that what you had in mind?

Yes! I think you're on the right track.

eHig:
I filled in the rest of the tests and fixed up a couple of mistakes. Let me know if it needs any more changes.

jdegoes:
Since one main purpose of this is to be able to use tree sets, I think picking a "least arbitrary" ordering for some things makes sense.

One can keep in mind the Haskell ordering for enums: `data A = B | C | D` (B comes before C comes before D, even though this is arbitrary, it is declaration order).

eHig:
I think the reference date comparison suggested by meets that criteria. For enums, they are compared by the case order in the schema. 

@thinkharderdev Is there anything else that needs changing?

thinkharderdev:
> I think the reference date comparison suggested by @thinkharderdev meets that criteria. For enums, they are compared by the case order in the schema. Is there anything else that needs changing?

I think it looks good. Looks like there are some test failures for the 2.12 build and the code needs formatting/linting but otherwise the implementation is great.

eHig:
oops, those should be fixed now

jdegoes:
@eHig @thinkharderdev What's the status of this one??

zio-schema/shared/src/main/scala/zio/schema/SchemaOrdering.scala
```diff
@@ -0,0 +1,87 @@
+package zio.schema
+
+import scala.annotation.tailrec
+
+import zio.Chunk
+import zio.schema.DynamicValue._
+import zio.schema.StandardType.UnitType
+
+object SchemaOrdering {
```
jdegoes, line 9: I might make all of this package private since the user should only ever have to use `Schema#ordering` to get access to this, and we can avoid cluttering the Scaladoc API.

eHig: this is updated now

eHig:
> @eHig @thinkharderdev What's the status of this one??

I think it should be ready to merge. Let me know if there are any changes needed

eHig:
thanks for your help!

PR #118, opened 2021-08-27
Fixes #31: Implement schema migrations

### LINKED ISSUE #31: Add intelligent transformation of data between two schemas
Given `Schema[A]`, and `Schema[B]`, we should be able to produce `Either[String, A => Either[String, B]]`, either alone, or with hints.

```scala
trait Schema[A] {
  ...
  def migrate[B](newSchema: Schema[B]): Either[String, A => Either[String, B]] = ???
}
```

A simple case is `Schema[B]` having smaller records and bigger enumerations, but the names of terms and relative position of structure is unchanged, in which case, the transition may always succeed.

Hints would allow more sophisticated transformations (including those requiring default values), as would an optimizer that can detect unambiguous field renames.

thinkharderdev:
This implements intelligent migrations between homomorphic schemas. The basic approach here is to use a tree diffing algorithm to diff the AST for two schemas. From the AST diff we get an "edit" script to transform an instance of one AST to another. Then we use the edit script to transform one value to another by way of a `DynamicValue`. 

Limitations of the current implementation:
1. There is no support for adding new nodes to a record. In the future we can use default values and/or annotations to provide "defaults".
2. Field/Case relabeling is not supported although the plumbing is there if we want to make it happen.


zio-schema/shared/src/main/scala/zio/schema/AstTransformation.scala
```diff
@@ -0,0 +1,330 @@
+package zio.schema
+
+import zio.{ Chunk, ChunkBuilder }
+
+sealed trait AstTransformation { self =>
```
jdegoes, line 5:
```suggestion
sealed trait Migration { self =>
```

???

zio-schema/shared/src/main/scala/zio/schema/AstTransformation.scala
```diff
@@ -0,0 +1,330 @@
...
+
+  def transformDynamic(value: DynamicValue): Either[String, DynamicValue] =
```
jdegoes, line 7:
```suggestion
  def migrate(value: DynamicValue): Either[String, DynamicValue] =
```

zio-schema/shared/src/main/scala/zio/schema/AstTransformation.scala
```diff
@@ -0,0 +1,330 @@
...
+        AstTransformation.tryIncrementDimension(value, path.toList, n)
+      case AstTransformation.DecrementDimensions(path, n) =>
+        AstTransformation.tryDecrementDimensions(value, path.toList, n)
+      case AstTransformation.UpdateFail(path, newMessage) =>
+        AstTransformation.tryUpdateFail(value, path.toList, newMessage)
+    }
+}
+
+object AstTransformation {
+  case class UpdateFail(path: Chunk[String], message: String) extends AstTransformation
```
jdegoes, line 29:
```suggestion
  final case class UpdateFail(path: Chunk[String], message: String) extends AstTransformation
```

Here and below. 👍

zio-schema/shared/src/main/scala/zio/schema/AstTransformation.scala
```diff
@@ -0,0 +1,330 @@
...
+
+  case class DeleteNode(path: Chunk[String]) extends AstTransformation
+
+  case class Relabel(path: Chunk[String], tranform: LabelTansformation) extends AstTransformation
+
+  case class IncrementDimensions(path: Chunk[String], n: Int) extends AstTransformation
+
+  case class DecrementDimensions(path: Chunk[String], n: Int) extends AstTransformation
+
+  def tryDerive(from: SchemaAst, to: SchemaAst): Either[String, Chunk[AstTransformation]] = {
```
jdegoes, line 47:
```suggestion
  def derive(from: SchemaAst, to: SchemaAst): Either[String, Chunk[AstTransformation]] = {
```

jdegoes (dismissed):
Superb bit of work! A few relatively minor comments. Will review more later.

zio-schema/shared/src/main/scala/zio/schema/Schema.scala
```diff
@@ -43,6 +43,18 @@ sealed trait Schema[A] {
    */
   def <+>[B](that: Schema[B]): Schema[Either[A, B]] = self.orElseEither(that)
 
+  /**
+   *  Convert to Schema[B] iff B and A are homomorphic.
+   *
+   *  This can be used to e.g convert between a case class and it's
+   *  "generic" representation as a ListMap[String,_]
+   */
+  def cast[B](newSchema: Schema[B]): Either[String, Schema[B]] =
```
jdegoes, line 52: Maybe `coerce` instead of `cast`?

thinkharderdev:
@jdegoes I think this guys is ready to go.

jdegoes (approved):

jdegoes:
Great to see this migration work land. 🎉

PR #126, opened 2021-09-28
Breakout Json codec into its own project

DamianReeves:


build.sbt
```diff
@@ -77,6 +76,28 @@ lazy val zioSchemaJS = zioSchema.js
 
 lazy val zioSchemaJVM = zioSchema.jvm
 
+lazy val zioSchemaJson = crossProject(JSPlatform, JVMPlatform)
```
thinkharderdev, line 79: We should aggregate `zioSchemaJsonJS` and `zioSchemaJsonJVM` under `root`. Right now `JsonCodecSpec` is not running in the CI job.

thinkharderdev (changes requested):
One small thing but looks great!

DamianReeves:
> One small thing but looks great!

done

PR #128, opened 2021-10-04
Rewrite macro derivation to use white box macro

thinkharderdev:
The basic problem we have with the Magnolia-based macro derivation is that the path-dependent `Accessors` type won't ever infer correctly. So to get various reified-optics code to compile users will either have to:

1. Create Schemas by hand 
2. Cast magnolia-derived Schemas to their correct type (i.e. `Schema.CaseClass3[...]`)

Neither is very ergonomic. 

We can solve the issue by replacing Magnolia with a white box def macro so that the type gets resolved after the macro expansion and everything infers correctly. 

This is an initial stab at the macro. Here we do a couple of things:

1. Pull most of the core stuff into a `zio-schema-core` subproject so we can have the def macro in a separate compilation unit (not sure we actually need to do this but that's the way it is for now).
2. Implement a whitebox macro for schema derivation. This is only implemented for Product types now (and only up to arity 3 but extending to all arities is trivial). 

What DOES NOT currently work:
1. Enum derivation.
2. Derivation for recursive types. 

Enums should be relatively straightforward but not sure how to handle recursive types.


thinkharderdev:
Returning this guy to draft. Will sort out the annotation capture and remove the magnolia derivation entirely and rebase this onto the project structure from https://github.com/zio/zio-schema/pull/143

thinkharderdev:
Rebased and resubmitted on a different branch

Issue #81, opened 2021-06-18
Patch value with Diff


thinkharderdev:
This is a placeholder to gather thoughts but the initial idea is to:

Once #78 is implemented, we can add a method to the `Schema[A]` trait:
```scala
trait Schema[A] { self =>
  def patch(diff: Diff): Either[String, A => Either[String,A]]
}
```

which returns a `Left(error message)` if the diff cannot be applied to the schema structure (e.g. if I try to apply a `Diff.Number` to a `Schema.Record` and otherwise returns a function which will apply the specified `diff` to value of type `A` (possibly failing). 

And associated syntax
```scala
implicit class DiffOps[A: Schema](a: A) {
   def patch(diff): Either[String,A] = Schema[A].patch(diff).flatMap(_.apply(a))
}
```

Laws:
1. Given `a: A` then `a.patch(Diff.Identical) == Right(a)`
2. Given `a1: A` and `a2: A` we should have `a1.patch(a1.diff(a2)) == Right(a2)`
3. Given `a1: A` and `a2: A` we should have 
```scala
val diff1 = a1.diff(a2)
val diff2 = a2.diff(a1)
a1.patch(diff1).flatMap(_.diff(diff2)) == Right(a1)
```

Some questions:
1. Should the patching be able to fail if the the structure is correct. That is, should the signature be `def patch(diff: Diff): Either[String, A => A]` instead?

dmitrykozinets:
I would like to give this one a try.

The 2 proposed signatures are:
def patch(diff: Diff): Either[String, A => Either[String,A]]
and
def patch(diff: Diff): Either[String, A => A]

If we have a type A when can the function fail going from A to A?

Not sure how to deal with String...  For example:
val diff1 = "abc" diffEach "xyaz"
which produces: Myers(Chunk(Insert(x),Insert(y),Keep(a),Insert(z),Delete(b),Delete(c)))

How do you apply diff1 to some other random String like "hello"?
Or should this fail when one of the steps (Insert, Keep or Delete) do not match and are not doable?

Any other thoughts or guidance would be appreciated as I am new to the Zio community.

thinkharderdev:
Thanks!

The way I was thinking about it is that the patch is intended to be applied to the original value so applying to completely different value creates a "merge conflict" which would fail.

[...]

PR #132, opened 2021-10-19
Patch value with Diff #81

dmitrykozinets:
This PR is for issue #81 
Disclaimer: I am new to ZIO, so any and all constructive criticism and feedback is most welcome 

I have left some TO DO comments, which I am not sure how to do.  Please take a look.


DamianReeves (approved):
LGTM

zio-schema/shared/src/main/scala/zio/schema/Diff.scala
```diff
@@ -254,6 +258,195 @@ object Differ {
...
+            case DynamicValue.Record(values) => values
+          }
+          patchProductData(structure.toChunk, values, diffs).asInstanceOf[Either[String, A]]
+        }
+
+      case (schema: Schema.Record[A], Diff.Record(diffs: ListMap[String, Diff])) =>
+        Right { (a: A) =>
+          val values: ListMap[String, DynamicValue] = schema.toDynamic(a) match {
+            case DynamicValue.Record(values) => values
+          }
```
thinkharderdev, line 395: Should probably do a wildcard case here just in case. It should in theory never happen but don't want to fail with a MatchError in case it does.

zio-schema/shared/src/main/scala/zio/schema/Diff.scala
```diff
@@ -254,6 +258,195 @@ object Differ {
...
+              case (Some(diff), None)                           => Some(Left(s"Value missing for diff=$diff."))
+              case (None, None)                                 => Some(Left(s"Unknown error in sequence."))
+            }
+            .partitionMap(identity) match {
+            case (Chunk.empty, values) => Right(fromChunk(values))
+            case (errors, _)           => Left(s"Running patch produced the following error(s): ${errors.toList}.")
+          }
+        }
+
+      // TODO - how do you tell if diff is from left or right?
```
thinkharderdev, line 370: I think we probably need to tag the original `Diff` with `Tag.Left` or `Tag.Right`.

thinkharderdev (commented):
Excellent work! Looks great overall. Once we sort out the last couple issues you mentioned then should be good to merge this. Thanks!

zio-schema/shared/src/main/scala/zio/schema/Diff.scala
dmitrykozinets: I created another Diff type to deal with eithers:
  `final case class Either(value: Diff, tag: Tag) extends Diff`

This is a little ugly, please let me know if you can think of a nicer way:
`      case (Schema.EitherSchema(leftSchema, _), Diff.Either(diff, Diff.Tag.Left)) =>
        patch(leftSchema, diff).map(
          f =>
            (a: A) =>
              Right(a.left.flatMap(f(_) match {
                case Right(value) => Left(value)
                case Left(value)  => Left(value)
              }))
        )
`
Thanks!

thinkharderdev: Ugly and working is better than pretty and not working :)

thinkharderdev (dismissed):
Looks great!

zio-schema/shared/src/main/scala/zio/schema/Diff.scala
```diff
@@ -254,6 +258,210 @@ object Differ {
         f.applyOrElse((thisValue, thatValue), (_: (A, A)) => Diff.NotComparable)
     }
 
+  def patch[A](schema: Schema[A], diff: Diff): Either[String, A => Either[String, A]] =
```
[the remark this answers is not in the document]
dmitrykozinets: @thinkharderdev Done

thinkharderdev (approved):

Issue #125, opened 2021-09-21
Support for Annotations on all types


dmitrykozinets:
Annotation support is needed on all types not just on Record.

Currently records support annotations, but it would be useful to also support annotations on all `Schema` types.

https://github.com/zio/zio-schema/blob/main/zio-schema/shared/src/main/scala/zio/schema/Schema.scala#L218

alexvanolst:
I'd like to pick this one up.

alexvanolst:
Other than Enum / Case is there anything left to do here? Happy to pick up the rest post hackathon if someone can point out any gaps.

dmitrykozinets:
@alexvanolst Thank you for picking this up!
We were thinking of adding annotation on all types and doing something like this:
```
sealed trait Schema[A] {
   def annotations: Chunk[Any]
}
```
@thinkharderdev @DamianReeves what do you think?

thinkharderdev:
Yes! I think we ultimately want to allow annotating any schema.

DamianReeves:
Agreed

alexvanolst:
I've opened a draft PR that gets it compiling with annotations added to the Schema trait.

Will probably need some feedback on what the desired behaviour is before it's merge-able since they don't do much right now.

thinkharderdev:
Closed by #147

PR #141, opened 2021-10-30
Support for Annotations on Enum / Case

alexvanolst:
Add support for Annotations on Enum / Case

Could use some feedback on

- How best to test this
- Not sure whether equals should be checking annotations
- What other types should I cover in a follow up to complete https://github.com/zio/zio-schema/issues/125


zio-schema/shared/src/main/scala/zio/schema/macros.scala
```diff
@@ -1317,22 +1317,25 @@ object DeriveSchema {
...
     type S = Subtype[Schema, A]#SType
+    val annotations = Chunk.fromIterable(sealedTrait.annotations)
     sealedTrait.subtypes.sortBy(_.typeName.short).map { subtype: Subtype[Schema, A] =>
       Schema.Case[S, A](
         id = subtype.typeName.short,
         codec = subtype.typeclass.asInstanceOf[Schema[S]],
         unsafeDeconstruct =
-          (a: A) => if (subtype.cast.isDefinedAt(a)) subtype.cast(a) else throw new IllegalArgumentException
+          (a: A) => if (subtype.cast.isDefinedAt(a)) subtype.cast(a) else throw new IllegalArgumentException,
+        annotations
```
thinkharderdev, line 1327: Here we want the annotations from the case class I think.

zio-schema/shared/src/test/scala/zio/schema/SchemaAssertions.scala
```diff
@@ -93,10 +93,10 @@ object SchemaAssertions {
       case (Schema.Tuple(left1, right1), Schema.Tuple(left2, right2)) =>
         equalsSchema(left1, left2) && equalsSchema(right1, right2)
       case (Schema.Optional(codec1), Schema.Optional(codec2))   => equalsSchema(codec1, codec2)
-      case (Schema.Enum1(l), Schema.Enum1(r))                   => equalsCase(l, r)
-      case (Schema.Enum2(l1, l2), Schema.Enum2(r1, r2))         => hasSameCases(Seq(l1, l2), Seq(r1, r2))
-      case (Schema.Enum3(l1, l2, l3), Schema.Enum3(r1, r2, r3)) => hasSameCases(Seq(l1, l2, l3), Seq(r1, r2, r3))
-      case (Schema.EnumN(ls), Schema.EnumN(rs))                 => hasSameCases(ls.toSeq, rs.toSeq)
+      case (Schema.Enum1(l, _), Schema.Enum1(r, _))                   => equalsCase(l, r) // TODO: Should equals compare annotations?
```
thinkharderdev, line 96: Yeah, we need to assert equality of the annotations here as well.

thinkharderdev (changes requested):
Looks great! Just a couple small comments. For testing, once we have `SchemaAssertions.equalsSchema` checking annotations are the same between operands then we can create a unit test in `DeriveSchemaSpec` similar to the "correctly derives schema for case class with nested case classes" spec.

zio-schema/shared/src/main/scala/zio/schema/macros.scala
thinkharderdev: So `subtype.annotations` instead of just `annotations`

zio-schema/shared/src/test/scala/zio/schema/SchemaAssertions.scala
alexvanolst: Will Chunk.equals work here?

thinkharderdev: Yeah, I think it should. It's fine for the test assertions anyway. If the annotations are not case classes then the reference equality won't work but as long as they are case classes it should work fine.

thinkharderdev:
@alexvanolst This looks good to merge. Just need to run `sbt prepare` to resolve the linter issues. Thanks for your work on this!

thinkharderdev (approved):
Great work!

Issue #138, opened 2021-10-22
Support native schema for Map


jdegoes:
Currently there is no native support for maps. However, some protocols and formats have a native representation for maps (e.g. DynamoDB). Consequently, there is a strong advantage to having a native map type, distinct from `Sequence`.

 - [ ] Add new Schema.Map case class, which stores a schema for key, stores a schema for value, and extends `Schema[Map[K, V]]`.
 - [ ] Support new schema type for all existing codecs and other operations that match on schema

@thinkharderdev Contender for ZIO Hackathon??

pierangeloc:
I'd like to work on this one

thinkharderdev:
> I'd like to work on this one

Awesome! Let me know if you have any questions!

PR #142, opened 2021-10-30
Provide direct support for Map

pierangeloc:
Resolves https://github.com/zio/zio-schema/issues/138

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ For the general guidelines, see ZIO [contributor's guide](https://github.com/zio
 Before you submit a PR, make sure your tests are passing, and that the code is properly formatted
 
 ```
-sbt test
+sbt prepare
 
-sbt fmt
+sbt test
 ```
--- a/README.md
+++ b/README.md
@@ -37,3 +37,17 @@ libraryDependencies += "dev.zio" %% "zio-schema" % "<version>"
 [link-discord]: https://discord.gg/2ccFBr4 "Discord"
 [Stage]: https://img.shields.io/badge/Project%20Stage-Development-yellowgreen.svg
 [Stage-Page]: https://github.com/zio/zio/wiki/Project-Stages
+
+## Contributing
+
+For the general guidelines, see ZIO [contributor's guide](https://github.com/zio/zio/blob/master/docs/about/contributing.md).
+
+#### TL;DR
+
+Before you submit a PR, make sure your tests are passing, and that the code is properly formatted
+
+```
+sbt test
+
+sbt fmt
+```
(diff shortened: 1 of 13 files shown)


pierangeloc:
@thinkharderdev While working on the Json serialization, and looking also at the Map as supported by DynamoDB (or by other formats such as Postgres hstore), it makes probably most sense to support Schema not for Map[K, V] with K and V having a Schema, but just for Map[String, V] with V having a Schema.
Should we apply this restriction?

thinkharderdev:
> @thinkharderdev While working on the Json serialization, and looking also at the Map as supported by DynamoDB (or by other formats such as Postgres hstore), it makes probably most sense to support Schema not for Map[K, V] with K and V having a Schema, but just for Map[String, V] with V having a Schema.
> Should we apply this restriction?

I think we do want to support `Map[K,V]`. For JSON serialization (and also for other codecs) we would encode this as an array of tuples. So something like
```scala
case class Key(keyName: String, data: Int)
case class Value(value: String)

case class Foo(map: Map[Key,Value]
```

would get JSON encoded as 
```
{"map":[{"left":{"keyName":"k","data":0},"right":{"value":"something"}}]
```

pierangeloc:
@thinkharderdev I implemented the codecs as you suggested, and added unit tests

thinkharderdev:
Looks great! I think you need to run Scalafmt again though to fix the listing errors.

pierangeloc:
@thinkharderdev thanks, rebased and reformatted!

thinkharderdev:
> @thinkharderdev thanks, rebased and reformatted!

Seems like there is another linter error

pierangeloc:
`build.sbt` was unformatted. I also added a small  paragraph for contributors in the readme.md

thinkharderdev:
> `build.sbt` was unformatted. I also added a small paragraph for contributors in the readme.md

Linter is still failing :). I think you want to run `sbt prepare` instead of `sbt fmt`. The latter will only run scalafmt while the former will run scalafix and scalafmt.

README.md
```diff
@@ -37,3 +37,17 @@ libraryDependencies += "dev.zio" %% "zio-schema" % "<version>"
...
+For the general guidelines, see ZIO [contributor's guide](https://github.com/zio/zio/blob/master/docs/about/contributing.md).
+
+#### TL;DR
+
+Before you submit a PR, make sure your tests are passing, and that the code is properly formatted
+
+```
+sbt test
+
+sbt fmt
```
thinkharderdev, line 52: I would reverse the order here and change `sbt fmt` to `sbt prepare`, so:
```
sbt prepare
sbt test
```
since on rare occassions scalafix can make changes that cause compilation to fail.

pierangeloc: @thinkharderdev thanks, now the linter should be happy 🤞

thinkharderdev (approved):
Awesome work!

PR #143, opened 2021-10-30
Basic Scala 3 compilation (no derivation)

DamianReeves:
Basic Scala 3 compilation (no derivation)


jdegoes:
This has aged, so will close for now, but supporting Scala 3 is very much a top priority for this project!

PR #145, opened 2021-11-01
Add native schema support for `Set`

### LINKED ISSUE #139: Support native schema for Set
Currently there is no native support for sets. However, some protocols and formats have a native representation for sets (e.g. Redis). Consequently, there is a strong advantage to having a native set type, distinct from `Sequence`.

 - [ ] Add new Schema.Set case class, which stores a schema for element type, and extends `Schema[Set[A]]`.
 - [ ] Support new schema type for all existing codecs and other operations that match on schema

@thinkharderdev Contender for ZIO Hackathon??

zalbia:
Resolves #139


thinkharderdev:
I think you linked the wrong issue on this one :). Should be https://github.com/zio/zio-schema/issues/138

PR #147, opened 2021-11-05
Support for annotations on all types

alexvanolst:
https://github.com/zio/zio-schema/issues/125

Adds `def annotations: Chunk[Any]` to `Schema[A]` implemented as a field on concrete Schema


thinkharderdev (commented):
This looks good in general. I don't think there is any specific use case from the library side for these annotations, we just want to allow users to capture arbitrary annotations for their own purposes. 

Two small changes I would make though:
1. Wherever we accept an annotations chunk in the constructor/method we should default it to `Chunk.empty` since in the modal case I think users will not use this and we shouldn't add the additional boilerplate. 
2. Add an `def annotate(annotation: Any): Schema[A]` method to the `Schema[A]` trait to make it easier to add ad-hoc annotations to derived Schemas.

zio-schema/shared/src/main/scala/zio/schema/Schema.scala
```diff
@@ -192,9 +194,9 @@ object Schema extends TupleSchemas with RecordSchemas with EnumSchemas {
   implicit val nil: Schema[Nil.type] = Schema[Unit].transform(_ => Nil, _ => ())
 
   implicit val none: Schema[None.type] = Schema[Unit].transform(_ => None, _ => ())
-
+  // TODO: I don't think it seems right to copy the annotations from `schemaA`/`element` here.
```
thinkharderdev, line 197: Agreed

zio-schema/shared/src/main/scala/zio/schema/Schema.scala
```diff
@@ -31,6 +31,8 @@ sealed trait Schema[A] {
 
   type Accessors[Lens[_, _], Prism[_, _], Traversal[_, _]]
 
+  def annotations: Chunk[Any]
+
```
thinkharderdev, line 35: I would add a method `def annotate(annotation: Any): Schema[A]` here

thinkharderdev (commented):
Thanks for your work on this! A couple of small suggestions but I think once we get the build green this should be good to go.

zio-schema/shared/src/main/scala/zio/schema/Schema.scala
alexvanolst: 👍 Will try find some time later today to do so.

thinkharderdev:
@dmitrykozinets @DamianReeves I'm happy with this implementation. Do you guys want to review it before merging?

thinkharderdev (approved):
Thanks for your work on this!

dmitrykozinets:
Looks good to me!  Thank you @alexvanolst !

Issue #77, opened 2021-06-10
Implement Thrift Codec


thinkharderdev:
We should implement a codec to serialize/deserialize a `Schema` to the Apache Thrift binary serialization format.

This should be added in the `zio-schema-thrift` subproject

fokot:
I will take it

thinkharderdev:
Closed by #163

PR #163, opened 2021-12-13
Thrift

fokot:
https://github.com/zio/zio-schema/issues/77


zio-schema-thrift/shared/src/main/scala/zio/schema/codec/ThriftCodec.scala
```diff
@@ -0,0 +1,1090 @@
...
+      v.foreach(encodeValue(None, schema, _))
+      p.writeListEnd();
+      fieldNumber.foreach(
+        _ => p.writeFieldEnd()
+      )
+
+    }
+
+
+    def encodeOptional[A](fieldNumber: Option[Short], codec: Schema[A], v: Option[A]): Unit =
```
thinkharderdev, line 235: With respect to both `Option` and `List` see the recent change to the Protobuf codec (https://github.com/zio/zio-schema/pull/157). We need to be able to encode `Option[Option[_]]` and `List[List[_]]` type structures so we may need to explicitly encode `None` so we can differentiate between `Some(None)` and `None` values. Likewise for `List(Nil)` and `Nil`. Thrift support lists of lists natively unlike protobuf so this may not be a problem but something to be aware of.

jdegoes:
Going to leave this one open since I know @fokot is still working on it!

zio-schema-thrift/shared/src/main/scala/zio/schema/codec/ThriftCodec.scala
fokot: This is a problem. I do not know how to do it. When I generate java classes via `thrift` command, they do it that way, that they omit null values, do not even encode the field name for it. I do not see how I can encode `Option[Option[Int]]` and differentiate between `None` and `Some(None)`. How to do it semantically? Thrift does not have a notion of `Option[_]`, only the `null` value. Nested lists, even empty ones are working as list always has the header with its size so when I have List[List[Int]] I can differentiate between `Nil` and `List[Nil]`. There is even test for it.

zio-schema-thrift/shared/src/main/scala/zio/schema/codec/ThriftCodec.scala
```diff
@@ -0,0 +1,1327 @@
...
+        new Decoder(ch)
+          .decode(Chunk.empty, schema)
+          .left
+          .map(
+            err => s"Error at path /${err.path.mkString(".")}: ${err.error}"
+          )
+
+  object Thrift {
+
+    def singleSchema[A](codec: Schema[A]): Schema[ListMap[String, _]] = Schema.record(Schema.Field("value", codec))
```
thinkharderdev, line 57: What is this for?

zio-schema-thrift/shared/src/main/scala/zio/schema/codec/ThriftCodec.scala
```diff
@@ -0,0 +1,1327 @@
...
+        Schema.Field("year", Schema.Primitive(StandardType.IntType)),
+        Schema.Field("month", Schema.Primitive(StandardType.IntType))
+      )
+
+    def durationStructure(): Seq[Schema.Field[_]] =
+      Seq(
+        Schema.Field("seconds", Schema.Primitive(StandardType.LongType)),
+        Schema.Field("nanos", Schema.Primitive(StandardType.IntType))
+      )
+  }
```
thinkharderdev, line 82: I think these can all be `val`s

zio-schema-thrift/shared/src/main/scala/zio/schema/codec/ThriftCodec.scala
```diff
@@ -0,0 +1,1327 @@
...
+
+  private def unwrapLazy(schema: Schema[_]) = schema match {
+    case Schema.Lazy(s) =>
+      s()
```
thinkharderdev, line 86: nit: In `Schema.Lazy` we assign the lazily derived `Schema` in a val for access without invoking the function again, so this can be changed to 

```case s @ Schema.Lazy(_) => s.schema```

File: zio-schema-thrift/shared/src/main/scala/zio/schema/codec/ThriftCodec.scala [hunk 5]
```diff
@@ -0,0 +1,1327 @@
...
+    def encodeSet[A](fieldNumber: Option[Short], schema: Schema[A], v: Set[A]): Unit = {
+      writeFieldBegin(fieldNumber, TType.SET)
+      p.writeSetBegin(new org.apache.thrift.protocol.TSet(getType(schema), v.size))
+      v.foreach(encodeValue(None, schema, _))
+    }
+
+    def encodeOptional[A](fieldNumber: Option[Short], codec: Schema[A], v: Option[A]): Unit =
+      v.foreach(encodeValue(fieldNumber, codec, _))
+
+    def encodeValue[A](fieldNumber: Option[Short], schema: Schema[A], value: A): Unit =
```
thinkharderdev, line 266: Can we add `//scalafmt: { maxColumn = 400, optIn.configStyleArguments = false }` for this method?

thinkharderdev (changes requested):
Looks awesome! A couple of small comments. As written this won't compile for Scala 2.12 since it uses a couple methods only available in 2.13. See https://github.com/fokot/zio-schema/pull/1/files for the changes needed to get it working on 2.12

zio-schema-thrift/shared/src/main/scala/zio/schema/codec/ThriftCodec.scala
fokot: Obviously not used. Deleted.

zio-schema-thrift/shared/src/main/scala/zio/schema/codec/ThriftCodec.scala
fokot: Changed

zio-schema-thrift/shared/src/main/scala/zio/schema/codec/ThriftCodec.scala
fokot: Great. Changed

zio-schema-thrift/shared/src/main/scala/zio/schema/codec/ThriftCodec.scala
fokot: :+1:

zio-schema-thrift/shared/src/main/scala/zio/schema/codec/ThriftCodec.scala
fokot: I explicitly encoded Options as in Protobuf. Plus decoder can decodes also omitted field as None (Thrift generated classes behavior).

thinkharderdev (approved):
Awesome work!

PR #170, opened 2021-12-27
Add schema for DynamicValue

### LINKED ISSUE #162: Add schema for DynamicValue
We should be able to use included codecs on `DynamicValue`s. This can be done by providing an implicit `Schema[DynamicValue]` in the library.

igrebenik:
Closes #162


igrebenik:
Hey @thinkharderdev!
I more or less figured out how to deal with `DynamicValue.Primitive[A]`. 
Basially `Primitive[A]` is limited by the number of `StandardType[A]` values, so I listed all Schema.Case for Primitive for which there are defined StandardType values.
However, I have no idea what to do with `DynamicValue.Singleton[A]` though...
@jdegoes Hey John! Maybe you have some thoughts on making schema for `DynamicValue.Singleton[A]`?

zio-schema/shared/src/main/scala/zio/schema/DynamicValue.scala
```diff
@@ -1866,4 +1871,592 @@ object DynamicValue {
...
+        (value, standardType) => DynamicValue.Primitive(value, standardType),
+        primitive => primitive.value,
+        primitive => primitive.standardType
+      ),
+      _.asInstanceOf[DynamicValue.Primitive[Unit]]
+    )
+
+  private val primitiveStringCase: Schema.Case[DynamicValue.Primitive[String], DynamicValue] =
+    Schema.Case(
+      "Primitive",
```
thinkharderdev, line 2093: We want the case IDs to be unique. So I think here we should use the primitive type name instead of "Primitive." That way we can use the simplest encoding. So I think something like:

```scala
Schema.Case(
  "String",
  Schema[String].transform(s => DynamicValue.Primitive(s, StandardType[String], _.value),
 _.asInstanceOf[DynamicValue.Primitive[String]
)
```

zio-schema/shared/src/main/scala/zio/schema/DynamicValue.scala
```diff
@@ -1866,4 +1871,592 @@ object DynamicValue {
...
+      "DynamicAst",
+      Schema.CaseClass1[SchemaAst, DynamicValue.DynamicAst](
+        Schema.Field("ast", SchemaAst.schema),
+        schemaAst => DynamicValue.DynamicAst(schemaAst),
+        dynamicAst => dynamicAst.ast
+      ),
+      _.asInstanceOf[DynamicValue.DynamicAst]
+    )
+
+//  private def singletonCase[A]: Schema.Case[DynamicValue.Singleton[A], DynamicValue] = ???
```
thinkharderdev, line 2066: For this one I think we just use a case for `Singleton[Unit]`. The singleton Schema isomorphic to `Unit` so when we turn a `DynamicValue` into an `A` using `Schema[A]` we get the singleton instance from the `Schema` itself, so there is nothing to encode in the `DynamicValue`. So I think this one can just be:

```scala
Schema.Case[DynamicValue.Singleton[Any], DynamicValue](
  "Unit",
  Schema[Unit].transform(_ => DynamicValue.Singleton(()), _ => ()),
 _.asInstanceOf[DynamicValue.Singleton[Any]]
)
```

thinkharderdev (commented):
Great work! Had a couple of comments and an approach that I think will work for the Singleton case. Thanks!

zio-schema/shared/src/main/scala/zio/schema/DynamicValue.scala
thinkharderdev: > Hey @thinkharderdev!
> I more or less figured out how to deal with `DynamicValue.Primitive[A]`.
> Basially `Primitive[A]` is limited by the number of `StandardType[A]` values, so I listed all Schema.Case for Primitive for which there are defined StandardType values.
> However, I have no idea what to do with `DynamicValue.Singleton[A]` though...
> @jdegoes Hey John! Maybe you have some thoughts on making schema for `DynamicValue.Singleton[A]`?

Awesome! Left a couple comments that I think should help address these points!

zio-schema/shared/src/main/scala/zio/schema/DynamicValue.scala
igrebenik: Oh I see!
For some reason I thought I need to avoid getting specific schemas implicitly)
This look so much better!
Thank you)

zio-schema/shared/src/main/scala/zio/schema/DynamicValue.scala
```diff
@@ -1866,4 +1871,419 @@ object DynamicValue {
...
+      "Double",
+      Schema.primitive[Double].transform(d => DynamicValue.Primitive(d, StandardType[Double]), _.value),
+      _.asInstanceOf[DynamicValue.Primitive[Double]]
+    )
+
+  private val primitiveBinaryCase: Schema.Case[DynamicValue.Primitive[Chunk[Byte]], DynamicValue] =
+    Schema.Case(
+      "Binary",
+      Schema.primitive[Chunk[Byte]].transform(ch => DynamicValue.Primitive(ch, StandardType[Chunk[Byte]]), _.value),
+      _.asInstanceOf[DynamicValue.Primitive[Chunk[Byte]]]
```
thinkharderdev, line 2147: These tests won't work because of type erasure. I think we need to do it like in here: https://github.com/thinkharderdev/zio-schema/commit/7077f7bd5b6c426b08cc76a24c986ee8be3a6e16

thinkharderdev (changes requested):

zio-schema/shared/src/main/scala/zio/schema/DynamicValue.scala
igrebenik: Yes, I got the same problem during testing
Will redo the casting according to your branch!

zio-schema/shared/src/main/scala/zio/schema/DynamicValue.scala
```diff
@@ -2087,203 +2087,260 @@ private[schema] object DynamicValueSchema { self =>
...
+      }
     )
 
   private val primitiveBinaryCase: Schema.Case[DynamicValue.Primitive[Chunk[Byte]], DynamicValue] =
     Schema.Case(
       "Binary",
-      Schema.primitive[Chunk[Byte]].transform(ch => DynamicValue.Primitive(ch, StandardType[Chunk[Byte]]), _.value),
-      _.asInstanceOf[DynamicValue.Primitive[Chunk[Byte]]]
+      Schema.primitive[Chunk[Byte]].transform(ch => DynamicValue.Primitive(ch, StandardType[Chunk[Byte]]), _.value), {
+        case dv @ DynamicValue.Primitive(_: Chunk[_], _) => dv.asInstanceOf[DynamicValue.Primitive[Chunk[Byte]]]
```
igrebenik, line 2163: I think it's safe to treat Chunk[Byte] as Chunk[_] since the information about Chunk[Byte] is going to be erased anyway

igrebenik: @thinkharderdev Should I squash all the commits? Unfortunately, because of some renaming in the `StandardType` I got a lot of changes in all those places where StandardType values are used

igrebenik:
@thinkharderdev Should I squash all the commits? Unfortunately, because of some renaming in the StandardType I got a lot of changes in all those places where StandardType values are used

thinkharderdev:
> @thinkharderdev Should I squash all the commits? Unfortunately, because of some renaming in the StandardType I got a lot of changes in all those places where StandardType values are used

They'll get squashed when the PR is merged.

thinkharderdev (approved):
Awesome work!

PR #173, opened 2022-01-08
Support native schema for Set #139

googley42:
This builds on @zalbia draft PR https://github.com/zio/zio-schema/pull/153 and resolves issue #139.

Direct set schema support is required by `zio-dynamodb` to support a native set codec for DynamoDB, and currently this issue is a blocker for us.


googley42:
so looks like the build is failing as I do not have an implementation for `Set` in [Diff](https://github.com/zio/zio-schema/blob/main/zio-schema/shared/src/main/scala/zio/schema/Diff.scala#L110) - any recommendations for an implementation @thinkharderdev ?

thinkharderdev:
> so looks like the build is failing as I do not have an implementation for `Set` in [Diff](https://github.com/zio/zio-schema/blob/main/zio-schema/shared/src/main/scala/zio/schema/Diff.scala#L110) - any recommendations for an implementation @thinkharderdev ?

Yes, the set diff would just be the union on the set of all elements on the left set no in the right set (tagged as `Left`) and the set of all elements in the right set not in the left set (tagged as `Right`). Let me know if that doesn't make sense. I think it should be pretty straightforward.

thinkharderdev (commented):
Thanks! This looks great. Once the diff is done we should be able to merge.

googley42:
hi @thinkharderdev , I have added Diff support for SetSchema - could you please review when you get some time

thinkharderdev (approved):
Awesome work!

PR #176, opened 2022-01-15
Diff refactor

thinkharderdev:
Refactor diff implementation. There are two main goals:

1. Add type parameter to `Diff`. This solves an otherwise very nasty issue in the Scala 3 support.

2. Use Least Common Subsequence (LCS aka Myers Diff) for all sequence diffs. This simplifies things a lot of unifies all collection diffs under a single, principled algorithm.


zio-schema/shared/src/main/scala/zio/schema/diff/LCSDiff.scala
```diff
@@ -0,0 +1,154 @@
+package zio.schema.diff
+
+import zio.schema.{ Diff, Differ }
+import zio.{ Chunk, ChunkBuilder }
+
+object LCSDiff {
```
jdegoes, line 6: Is it possible to make this package private? Or, seeing as how the `apply` method builds a `Differ`, maybe we should place it into the companion object of `Differ`, where developers will look for constructors of `Differ`?

zio-schema/shared/src/main/scala/zio/schema/Diff.scala
```diff
@@ -413,427 +419,324 @@ object Differ {
...
-      override def apply(thisValue: A, thatValue: A): Diff =
-        f.applyOrElse((thisValue, thatValue), (_: (A, A)) => Diff.NotComparable)
+      override def apply(thisValue: A, thatValue: A): Diff[A] =
+        f.applyOrElse((thisValue, thatValue), (_: (A, A)) => Diff.notComparable)
     }
 
 }
 
-sealed trait Diff { self =>
+sealed trait Diff[A] { self =>
```
jdegoes, line 458: I quite like the addition of this type parameter. And we don't lose anything, really, since the "untyped" version is just `Diff[_]`. ❤️ 👍

jdegoes (approved):

Issue #193, opened 2022-02-16
Allow coercions during DynamicValue deserialization


jdegoes:
When attempting to deserialize a value of type `A` from a `DynamicValue`, we should be flexible in allowing reading "smaller" types from "larger" types, i.e. allow coercions.

Various examples:

1. Allow coercing the string `"123"` to the number `123`.
2. Allow coercing the integer `123` to the floating-point number `123.0F`.
3. Etc.

All these coercions would be (for now) on primitive types. Therefore, one way to implement this may be to modify `StandardType`, to add something like:

```scala
sealed trait StandardType[A] {
  def coerce[B](that: StandardType[B]): Option[A => Either[String, B]]
}
```

If the types cannot be coerced under any circumstance, then `coerce` would return `None`. For example, it is not possible to coerce a `Boolean` into a `java.time.Instant`. On the other hand, if types can be coerced under some circumstances, depending on their values, then `coerce` can return `Some`, which will hold the conversion function `A => Either[String, B]`.

The conversion function may fail, because even though some values may be coercible, other values may not be coercible. For example, strings can sometimes be coerced to numbers, but not always, because not all strings contain numbers.

By defining and placing such a coercion method on `StandardType`, and leaving it _abstract_, it will motivate implementing it on the standard type subtypes, for all types that make sense.

Then, after such coercion machinery is in place, it should be relatively straightforward to support coercions when deserializing from `DynamicValue`.

@thinkharderdev Any other places it can make sense to use coercion? How about during migrations?

PR #197, opened 2022-02-16
Add `coerce` to StandardType

kitlangton:
Working on #193

### DIFF
--- a/tests/shared/src/test/scala/zio/schema/StandardTypeCoerceSpec.scala
+++ b/tests/shared/src/test/scala/zio/schema/StandardTypeCoerceSpec.scala
@@ -17,8 +17,8 @@ object StandardTypeCoerceSpec extends DefaultRunnableSpec {
           assertTrue(
             coercion.get("-1") == Right(-1),
             coercion.get("1") == Right(1),
-            coercion.get("hello") == Left("String hello could not be coerced into an Int"),
-            coercion.get("1.5") == Left("String 1.5 could not be coerced into an Int")
+            coercion.get("hello").isLeft,
+            coercion.get("1.5").isLeft
           )
         },
         test("-> Long") {
@@ -28,8 +28,8 @@ object StandardTypeCoerceSpec extends DefaultRunnableSpec {
             coercion.get("-5") == Right(-5L),
             coercion.get("5") == Right(5L),
             coercion.get("999999999999") == Right(999_999_999_999L),
-            coercion.get("hello") == Left("String hello could not be coerced into a Long"),
-            coercion.get("1.5") == Left("String 1.5 could not be coerced into a Long")
+            coercion.get("hello").isLeft,
+            coercion.get("1.5").isLeft
           )
         },
         test("-> Float") {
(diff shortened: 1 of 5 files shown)


zio-schema/shared/src/main/scala/zio/schema/StandardType.scala
```diff
@@ -99,6 +105,135 @@ object StandardType {
...
+                .toRight(s"String $a could not be coerced into a BigInteger")
+          )
+        case DayOfWeekType =>
+          Some(
+            a =>
+              scala.util.Try(DayOfWeek.valueOf(a)).toOption.toRight(s"String $a could not be coerced into a DayOfWeek")
+          )
+        case MonthType =>
+          Some(
+            a => scala.util.Try(Month.valueOf(a)).toOption.toRight(s"String $a could not be coerced into a Month")
```
jdegoes, line 152: For performance reasons, maybe good to have an `EitherUtil.attemptOrElse[A](a: => A, msg: E): Either[E, A]`.

kitlangton:
@jdegoes One question I have is how _coerce-happy_ do we want to be? Should I be coercing strings to all possible types? Should I be coercing doubles to ints? Or should we solely coerce when we don't lose any information? Given the examples in the issue, I've erred on the side of coercing everything (_Javascript-style_).

zio-schema/shared/src/main/scala/zio/schema/StandardType.scala
```diff
@@ -99,6 +108,112 @@ object StandardType {
     override def tag                                  = Tags.STRING
     override def compare(x: String, y: String): Int   = x.compareTo(y)
     override def defaultValue: Either[String, String] = Right("")
+
+    override def coerce[B](that: StandardType[B]): Option[String => Either[String, B]] =
+      that match {
+        case UnitType   => None
+        case StringType => Some(a => Right(a))
+        case BoolType   => Some(a => a.toBooleanOption.toRight(s"String $a could not be coerced into a Boolean"))
```
jdegoes, line 116: Consider quoting the strings in all these cases, or you will get confusing error messages if the string has space characters.

e.g. if $a = "could not be", then the error message would be:

```
String could not be could not be coerced into a Boolean
```

But if you quote it:

```
String "could not be" could not be coerced into a Boolean
```

tests/shared/src/test/scala/zio/schema/StandardTypeCoerceSpec.scala
```diff
@@ -0,0 +1,169 @@
...
+            coercion.get("PT1S") == Right(java.time.Duration.of(1, temporalUnit)),
+            coercion.get("hello") == Left("String hello could not be coerced into a Duration")
+          )
+        },
+        test("-> Period") {
+          val coercion = StandardType.StringType.coerce(StandardType.PeriodType)
+
+          assertTrue(
+            coercion.get("P1Y2M3D") == Right(java.time.Period.of(1, 2, 3)),
+            coercion.get("hello") == Left("String hello could not be coerced into a Period")
```
jdegoes, line 115: I suggest only testing that it `isLeft`, because that makes the tests less brittle and the error messages easier to improve.

thinkharderdev:
> @jdegoes One question I have is how _coerce-happy_ do we want to be? Should we be coercing strings to all possible types? Should we be coercing doubles to ints? Or should we solely coerce when we don't lose any information? Given the examples in the issue, I've erred on the side of coercing everything (_Javascript-style_).

Maybe we can differentiate between those cases. So maybe we have coerce and coerceLossy?

PR #198, opened 2022-02-16
Begin to implement field matching

oliveiosello:
This is still incomplete in that I am currently handling cases of basic ambiguity and fuzzy matching, but not type coercion or accounting for Levenstein distances. However, I am hoping for some general feedback as well as some more complex examples of ambiguity errors. Thank you!


oliveiosello:
@jdegoes Howdy! I would be thankful for a review of this PR. As mentioned in the submission comment, it's still incomplete, but I think feedback would be beneficial before moving forward.

zio-schema/shared/src/main/scala/zio/schema/Schema.scala
```diff
@@ -152,6 +152,45 @@ sealed trait Schema[A] {
 }
 
 object Schema extends TupleSchemas with RecordSchemas with EnumSchemas {
+
+  private def fuzzyReformat(str: String): String = str.replaceAll("[-_]", "").toLowerCase
```
thinkharderdev, line 156: nit: `String.replaceAll` will compile the regex on every invocation. Probably want to define a static regex and reuse it. So something like:

```scala
val reformatPattern = Pattern.compile("[-_]")

def fuzzyReformat(str: String): String = reformatPattern.matcher(str).replaceAll("")
```

zio-schema/shared/src/main/scala/zio/schema/Schema.scala
```diff
@@ -152,6 +152,45 @@ sealed trait Schema[A] {
...
+  def bestFieldMatches(
+    record: Schema.Record[_],
+    pairs: Set[(String, StandardType[_])]
+  ): Either[String, Map[String, Int]] =
+    pairs.foldLeft[Either[String, Map[String, Int]]](Right(Map.empty)) {
+      case (acc, (name, valueType)) =>
+        for {
+          acc0  <- acc
+          index <- bestFieldMatch(record, name, valueType).toRight(s"No field found for $name")
+          _     <- if (acc0.values.exists(_ == index)) Left(s"Field $name is ambiguous") else Right(())
```
thinkharderdev, line 190: I think here we probably only want to consider the match ambiguous if there are multiple exact matches or multiple fuzzy matches, but an exact match should "pre-empt" a fuzzy match. We'll also eventually need a transform function to map the field names in cases of fuzzy matches (so we can apply migrations). So maybe we can define:

```scala
import zio.schema.ast.LabelTransformation

sealed trait LabelMatch
object LabelMatch {
  case class Exact(index: Int) extends LabelMatch
  // Here 
  case class Fuzzy(index: Int, fn: LabelTransformation) extends LabelMatch
}
```

zio-schema/shared/src/main/scala/zio/schema/Schema.scala
```diff
@@ -152,6 +152,45 @@ sealed trait Schema[A] {
 }
 
 object Schema extends TupleSchemas with RecordSchemas with EnumSchemas {
+
+  private def fuzzyReformat(str: String): String = str.replaceAll("[-_]", "").toLowerCase
+
+  @tailrec
+  private def isTypeMatch(schema: Schema[_], valueType: StandardType[_]): Boolean =
```
thinkharderdev, line 159: Should be able to leverage the schema equality work recently added by @vigoo here.

thinkharderdev (commented):
Thanks for your work on this! A few comments but I think this is on the right track. Apologies for the delayed review!

jdegoes:
This code has bit-rotted please feel free to re-open if you want to continue working!

PR #204, opened 2022-02-20
Unify schema of Either/Left/Right

### LINKED ISSUE #184: Codecs for Nothing, Either, Left, Right
Currently there is no implicit schema for `Nothing` which makes it hard to write generic functions requiring an implicit schema for example for an `Either[E, A]` type.
Defining our own schema (with `Schema.fail`) for that shows a next issue, that the codec derived by the schema for `Left` and `Right` is incompatible with the codec derived for the schema `Either[Nothing, A]` or `Either[A, Nothing]`. I think these should be equivalent.

vigoo:
Resolves #184


thinkharderdev (approved):
nice!

Issue #192, opened 2022-02-16
Dynamic schemas for payloads


jdegoes:
Consider the following use cases: cases where the user-defined data structure stores a "payload".

There are two classes of payload storage:

1. **Dynamic payload**. For example, an event could have an attached JSON blob, which stores details the user's application is not aware of, but which are important to capture and preserve.
2. **Semi-dynamic payload**. For example, a user data structure is generic, and can store any `A` for which there exists a suitable Schema. The task, then, is to describe the schema of the generic user data structure, which itself stores a schema, and a value that is described by the schema.

The ability to support (2) would be enabled by a `flatMap` like operation, which would allow you to define a schema dynamically. However, given schema is invariant and we wish to support operations that demand this invariance, adding such an ability is impossible.

Something more constrained may be useful:

1. In the first case, we can add `Schema.Dynamic`, a `case object` that extends `Schema[DynamicValue]`. This is truly a dynamic schema, in the sense that it describes the full range of possible values supported by ZIO Schema--a true superset of JSON and many other common self-describing semi-structured data formats. This data type can be supported fully or partially for many use cases: for example, we can always support `Schema[DynamicValue]` for JSON, or for relational databases that support JSON, etc. In protocols that support nested data and unions, we can emulate `DynamicValue` with a big union type. Some protocols may not have a good way to encode `DynamicValue` (e.g. a relational database table without any JSON columns). However, this is already true of some schemas, which describe data that cannot be encoded easily into some format.
2. In the second case, we don't need (or even want) to store any possible value. Rather, we would like to constrain the types of values we store by a specified schema. In this case, a schema that can describe the structure of both an `A` together with its schema, could be necessary and sufficient for solving the problem.

```scala
object Schema {
  ...
  implicit case object Dynamic extends Schema[DynamicValue]

  final case class SemiDynamic[A](schema: Schema[A]) extends Schema[(A, Schema[A]]
}
```

The latter `SemiDynamic` represents a promise that, given a schema of type `A`, any codec could serialize or deserialize both the schema, as well as a value of that type.

Together, these two new additions to schema should allow very flexible handling of dynamic and semi-dynamic payloads.

/cc @vigoo

PR #205, opened 2022-02-20
Dynamic and SemiDynamic schemas

vigoo:
- Resolves https://github.com/zio/zio-schema/issues/192
- Lots of unrelated Protobuf and Thrift codec fixes
- `Diff.fromSchema` is not implemented for `Dynamic` / `SemiDynamic` in this PR


zio-schema-protobuf/shared/src/main/scala/zio/schema/codec/ProtobufCodec.scala
```diff
@@ -724,6 +754,24 @@ object ProtobufCodec extends Codec {
         case StandardType.LongType   => varIntDecoder
         case StandardType.FloatType  => floatDecoder
         case StandardType.DoubleType => doubleDecoder
+        case StandardType.BigIntegerType =>
+          binaryDecoder.map { bytes =>
+            new java.math.BigInteger(bytes.toArray)
+          }
+        case StandardType.BigDecimalType =>
```
thinkharderdev, line 761: **facepalm**

zio-schema/shared/src/main/scala/zio/schema/Diff.scala
```diff
@@ -315,6 +316,8 @@ object Differ {
     case Schema.Enum21(c1, c2, c3, c4, c5, c6, c7, c8, c9, c10, c11, c12, c13, c14, c15, c16, c17, c18, c19, c20, c21, _)      => enumN(c1, c2, c3, c4, c5, c6, c7, c8, c9, c10, c11, c12, c13, c14, c15, c16, c17, c18, c19, c20, c21)
     case Schema.Enum22(c1, c2, c3, c4, c5, c6, c7, c8, c9, c10, c11, c12, c13, c14, c15, c16, c17, c18, c19, c20, c21, c22, _) => enumN(c1, c2, c3, c4, c5, c6, c7, c8, c9, c10, c11, c12, c13, c14, c15, c16, c17, c18, c19, c20, c21, c22)
     case Schema.EnumN(cs, _)                                                                                                   => enumN(cs.toSeq: _*)
+    case Schema.Dynamic(_)                                                                                                     => ??? // TODO
+    case Schema.SemiDynamic(_, _)                                                                                              => ??? // TODO
```
thinkharderdev, line 320: Not quite sure how to handle diffs on `Dynamic` but for `SemiDynamic` I think it should be relatively straightforward. It would just be a diff on the values (using the packaged schema).

zio-schema/shared/src/main/scala/zio/schema/Schema.scala
```diff
@@ -488,6 +491,38 @@ object Schema extends TupleSchemas with RecordSchemas with EnumSchemas with Sche
...
+      Right(DynamicValue.NoneValue)
+
+    /**
+     * Returns a new schema that with `annotation`
+     */
+    override def annotate(annotation: Any): Schema[DynamicValue] =
+      this.copy(annotations = annotations :+ annotation)
+
+    override def makeAccessors(b: AccessorBuilder): Unit = ()
+  }
```
thinkharderdev, line 510: Not sure I follow what the purpose of this is. `DynamicValue` is already an ADT which has an implicit schema associated with it.

thinkharderdev (commented):
Seems reasonable, but I'm not sure I understand what `Schema.Dynamic` is doing or why we need it.

zio-schema/shared/src/main/scala/zio/schema/Schema.scala
vigoo: I think what this gives us is to handle dynamic values specially in codecs. But I did not take any advantage of this in this PR.

So in theory for example in JSON you could probably encode a dynamic value as a JSON subtree without encoding it with the more verbose dynamic value ADT. 
But maybe that would loose some important type information and not possible anyway so in this case this really does not make much sense.

I'm ok with removing it, but I think conceptually it is nice to have it as a special schema, as described in the linked issue (https://github.com/zio/zio-schema/issues/192)

zio-schema/shared/src/main/scala/zio/schema/Diff.scala
vigoo: Sure I just wanted to have working `SemiDynamic` serialization asap so I can continue with zio-flow and defer this to a later PR if possible.

zio-schema/shared/src/main/scala/zio/schema/Schema.scala
thinkharderdev: @jdegoes 

It makes sense from the standpoint of encoders, I'm just concerned that it ends up creating weird edge cases and unexpected error conditions because `Dynamic` doesn't really describe anything so there is no way to handle, for example, diffing. And without a `Schema` `DynamicValue` doesn't seem like a particularly useful construct. If we want to handle arbitrary JSON, could we just have a `Schema` for the `zio.json.ast.Json`? And then maybe a constructor for `DynamicValue` that takes a JSON AST?

jdegoes:
First off, the context: ZIO Flow requires the ability to serialize schemas for user-defined data types. It needs to know the structure of those data types. So schemas being themselves transmitted over the wire, and stored and read, is important.

`Dynamic` describes the schema of semi-structured data (i.e. all data that can be stored in `DynamicValue`).

I'm not 100% convinced that we need to specialize `DynamicValue`  (since we do have a schema for it already), but there are a few advantages:

1. For serialization, there is no need to store the gigantic sum schema of the current implicit schema. Also, the current schema has no information that this gigantic sum represents semi-structured data. Of course, you can reverse engineer the information.
2. Many formats have special treatment for unstructured data. For example, JSON can store JSON values directly; relational databases would prefer to encode semis-structured data using a JSON column type; some databases might have to store them using a serialized string. 

Of course, there is high cost for specializing any schema. So a few pros do not mean the answer is to add it. 

The situation is different for SemiDynamic, I think, because we don't have a workaround, as far as I can see: which is describing the schema of both a schema, together with a value of that type.

thinkharderdev:
> First off, the context: ZIO Flow requires the ability to serialize schemas for user-defined data types. It needs to know the structure of those data types. So schemas being themselves transmitted over the wire, and stored and read, is important.
> 
> `Dynamic` describes the schema of semi-structured data (i.e. all data that can be stored in `DynamicValue`).
> 
> I'm not 100% convinced that we need to specialize `DynamicValue` (since we do have a schema for it already), but there are a few advantages:
> 
> 1. For serialization, there is no need to store the gigantic sum schema of the current implicit schema. Also, the current schema has no information that this gigantic sum represents semi-structured data. Of course, you can reverse engineer the information.
> 2. Many formats have special treatment for unstructured data. For example, JSON can store JSON values directly; relational databases would prefer to encode semis-structured data using a JSON column type; some databases might have to store them using a serialized string.
> 
> Of course, there is high cost for specializing any schema. So a few pros do not mean the answer is to add it.
> 
> The situation is different for SemiDynamic, I think, because we don't have a workaround, as far as I can see: which is describing the schema of both a schema, together with a value of that type.

Yeah, `SemiDynamic` makes sense but I'm just worried that adding a schema that acts as a sort of phantom type will cause problems elsewhere. One possibility would be to tweak `SemiDynamic` to implement `Schema[(A,Option[Schema[A]]]` so it can either be a "self-describing" datatype or something entirely dynamic. The you can fold the encoder/decoder logic for `Dynamic` into `SemiDynamic`. That is, if the payload has a schema, use that the decode the value, otherwise decode to `DynamicValue`. 

I think in principle we should also actually be able to infer a`Schema` from a `DynamicValue` which also opens some interesting possibilities.

jdegoes:
> Yeah, SemiDynamic makes sense but I'm just worried that adding a schema that acts as a sort of phantom type will cause problems elsewhere.

How is this different than `Schema[Int]`? In fact, maybe `DynamicValue` should be a `StandardType`. That would be a nice place to put it, if we decide to have `Schema[DynamicValue]` directly. JSON or semi-structured data is indeed a primitive in some systems.

thinkharderdev:
> > Yeah, SemiDynamic makes sense but I'm just worried that adding a schema that acts as a sort of phantom type will cause problems elsewhere.
> 
> How is this different than `Schema[Int]`? In fact, maybe `DynamicValue` should be a `StandardType`. That would be a nice place to put it, if we decide to have `Schema[DynamicValue]` directly. JSON or semi-structured data is indeed a primitive in some systems.

It is similar in the sense that the schema itself does not have any information about the structure of the data type, but the difference is that it is not obvious how we would implement various operations on `DynamicValue` where it is for `Int`. For example, how do we define an `Ordering` on `DynamicValue`?

jdegoes:
> For example, how do we define an Ordering on DynamicValue?

Ordering is already defined on DynamicValue through its existing Schema. DynamicValue is a sum type and is therefore ordered by its terms, and then within each term (products), by their terms, etc., recursively. I think all operations make sense for it.

thinkharderdev:
> > For example, how do we define an Ordering on DynamicValue?
> 
> Ordering is already defined on DynamicValue through its existing Schema. DynamicValue is a sum type and is therefore ordered by its terms, and then within each term (products), by their terms, etc., recursively. I think all operations make sense for it.

Yeah, fair enough.

thinkharderdev (approved):

PR #206, opened 2022-02-24
Equality implementation for Schema

vigoo:
This PR includes:
- strict equality function for `Schema[_]`
- structural equality function for `Schema[_]` which ignores transformations (both `Transform` and `Sequence`)
- fix for recursive `SchemaAst` materialization (tests before my new equality functions did not catch that it resulted in endless recursion)
- New cases for `SchemaAst` explicitly representing `Tuple` and `Either`, so we don't loose the distinction 
- Migration between products and tuples and sums and eithers

This is part of the required changes for https://github.com/zio/zio-schema/pull/205 to work.


zio-schema/shared/src/main/scala/zio/schema/SchemaEquality.scala
```diff
@@ -0,0 +1,76 @@
...
+            l.annotations == r.annotations && lEnum.structure === rEnum.structure
+          case (lRecord: Schema.Record[_], rRecord: Schema.Record[_]) =>
+            l.annotations == r.annotations && lRecord.structure === rRecord.structure
+          case (lMap: Schema.MapSchema[_, _], rMap: Schema.MapSchema[_, _]) =>
+            lMap.annotations == rMap.annotations && lMap.ks === rMap.ks && lMap.vs === rMap.vs
+          case (lSet: Schema.SetSchema[_], rSet: Schema.SetSchema[_]) =>
+            lSet.annotations == rSet.annotations && lSet.as === rSet.as
+          case (lSeq: Schema.Sequence[_, _], rSeq: Schema.Sequence[_, _]) =>
+            (lSeq.fromChunk eq rSeq.fromChunk) &&
+              (lSeq.toChunk eq rSeq.toChunk) &&
```
thinkharderdev, line 32: Wouldn't this effectively just turn schema equality into reference equality (at least for `Sequence`)?

zio-schema/shared/src/main/scala/zio/schema/SchemaEquality.scala
```diff
@@ -0,0 +1,76 @@
...
+              true
+            else {
+              if (visitedLazy.contains(lLazy) && visitedLazy.contains(rLazy))
+                true
+              else {
+                recursiveEqual(lLazy.schema, rLazy.schema, visitedLazy + lLazy + rLazy)
+              }
+            }
+          case (lLazy: Schema.Lazy[_], r: Schema[_]) =>
+            recursiveEqual(lLazy.schema, r, visitedLazy + lLazy)
```
thinkharderdev, line 69: Ahh, very clever. Much better than what I came up with :)

thinkharderdev (commented):
Awesome work! Had one comment but maybe I am just misunderstanding something

zio-schema/shared/src/main/scala/zio/schema/SchemaEquality.scala
vigoo: Those are conversion functions like in `Transform`. Probably that `identity` trick should be applied for sequence too, or fixed into `Chunk` and other collection types encoded as a combination of `Sequence` and `Transform`

vigoo: I first wanted to just get rid of the conversion and use `Transform` everywhere but it made hard to create the traversal optics so I just did the same extra identity field as with transform.

jdegoes:
@vigoo I'd suggest a couple revisions:

1. Rather than solving the "transform" problem twice, we can simplify `Sequence` so it always materializes a `Chunk` (it will extend `Schema[Chunk[A]]`), and then implement the schemas for other collection types using `Transform`. Probably, this was always a better way for this code to be written.
2. I'd suggest that using a LighweightTypeTag (from `Izumi Reflect`) for Transform may work well. Or possibly, its string representation. It's strongly connected to the target type of the transform node, and when combined with (1), would result in an identity for `Schema[List[A]]` as `scala.collections.immutable.List[<A>]`. 
3. If source code locations are used as identities, column numbers should be included, as row/col isn't always sufficient; e.g. `{ schema1(); schema2(); schema3() }` (on a single line, in a single file).

PR #211, opened 2022-02-27
adding Enum.structureWithAnnotations to fix #210

### LINKED ISSUE #210: `Enum`.`structure` is losing annotations of its cases
when calling method `structure` on an `Enum` the annotations of its `Case``s are not present in the structure anymore.
expected was, that annotations of its cases are preserved

```scala
          val caseA = Schema.Case[String, String]("A", Schema.primitive(StandardType.StringType), identity)
          val caseB = Schema.Case[String, String]("B", Schema.primitive(StandardType.StringType), identity)
          val caseC = Schema.Case[String, String]("C", Schema.primitive(StandardType.StringType), identity)
          val caseCWithAnnotation = caseC.copy(annotations = caseC.annotations + "Keep me!")
          val schema = Schema.Enum3(caseA, caseB, caseCWithAnnotation)

          val x = schema.structure
// x(2)._2.annotations.size = 0 but expected >=1
```

TobiasPfeifer:
instead of breaking the current API of `Enum.structure` this is adding `Enum.structureWithAnnotations` to enable annotation processing on enum cases.
fixes #210


zio-schema/shared/src/main/scala/zio/schema/Schema.scala
```diff
@@ -512,8 +513,8 @@ sealed trait EnumSchemas { self: Schema.type =>
 
     override def makeAccessors(b: AccessorBuilder): b.Prism[Z, A] = b.makePrism(self, case1)
 
-    override def structure: ListMap[String, Schema[_]] =
-      ListMap(case1.id -> case1.codec)
+    override def structureWithAnnotations: ListMap[String, (Schema[_], Chunk[Any])] =
+      ListMap(case1.id -> (case1.codec -> case1.annotations))
```
thinkharderdev, line 517: Should these just get folded into the schema itself? So something like:

```
override def structure: ListMap[String, Schema[_]] =
       ListMap(case1.id -> case1.codec.withAnnotations(case1.annotations))
```

TobiasPfeifer: Why are there annotations on a `Case` at all then?

I was concerned with annotation conflicts or semantic changes when folding down the annotations onto the inner codec. For example:
```
          case class name(name: String) extends StaticAnnotation
          val someRecord = Schema.record(???)
          val someRecordWithNameAnnotation = someRecord.annotate(name("SomeRecord"))
          val inner = Schema.Enum1(Case("Inner", someRecordWithNameAnnotation, ???, ???), Chunk(name("MyInner"))).annotate(name("MyInnerEnum"))
          val outer = Schema.Enum1(Case("Outer", inner, ???, ???), Chunk(name("MyOuter"))).annotate(name("MyOuterEnum"))


          @name("SomeRecord") case class SomeRecord(s: String)

          @name("MyInnerEnum") sealed trait InnerEnum
          @name("MyInner") case class Inner(someRecord: SomeRecord) extends InnerEnum

          @name("MyOuterEnum") sealed trait OuterEnum
          @name("MyOuter") case class Outer(inner: InnerEnum) extends OuterEnum
```

With nested enums representing a sealed trait hierarchy, I would not expect the `name` annotation to be pushed down (especially when handling the cases recursively). Also, this would lead to annotation conflicts.

thinkharderdev: Ok, fair enough. It is a bit awkward to have annotations on the `Case` itself so this seems like a reasonable approach.

thinkharderdev (approved):

PR #220, opened 2022-03-03
Add special encoding for List and Map types in SchemaAst

thinkharderdev:
Fixes issue with `SchemaAst` encoding found by the eagle-eyed @vigoo . Because we were using attributes `dimension` and `optional` on AST nodes to encode `List` and `Option` we had no way of distinguishing `List[Option[T]]` from `Option[List[T]]` when materializing. This PR adds a special `ListNode` AST node to encode `Schema.Sequence` and also `Dictionary` to encode `Schema.MapSchema`. 

Today's lesson in "I should have known better" is to not get too clever with encodings because it will come back to bite you! 

Also found that `DynamicValue.toTypedValue` was not handling `DynamicValue.Dictionary` so fixed that as well. 

The test coverage on Schema migrations is pretty lacking at the moment. Working on a separate PR to add a more robust set of test cases.

### DIFF
--- a/project/BuildHelper.scala
+++ b/project/BuildHelper.scala
@@ -82,7 +82,9 @@ object BuildHelper {
       if (sys.env.contains("CI")) {
         Seq("-Xfatal-warnings", "-Ypatmat-exhaust-depth", "80")
       } else {
-        Seq("-Ypatmat-exhaust-depth", "80")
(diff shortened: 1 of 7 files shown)


vigoo (approved):
Looks good!

PR #236, opened 2022-04-05
Add SemiDynamic/Dynamic to the structural equality check

vigoo:


zio-schema/shared/src/main/scala/zio/schema/SchemaEquality.scala
```diff
@@ -80,6 +80,10 @@ trait SchemaEquality {
             recursiveEqual(lTransform.codec, r, visitedPairs)
           case (l: Schema[_], rTransform: Schema.Transform[_, _, _]) if ignoreTransformations =>
             recursiveEqual(l, rTransform.codec, visitedPairs)
+          case (l: Schema.SemiDynamic[_], r: Schema.SemiDynamic[_]) =>
+            l.annotations == r.annotations
```
thinkharderdev, line 84: Should we compare the default value here instead of just the annotations?

thinkharderdev (commented):
Sorry for the delay, had a lot of travel this week

jdegoes:
@vigoo I am not sure if this is still relevant. Please re-open if so! 🙏

Issue #182, opened 2022-01-30
decoder of JsonCodec and ProtobufCodec aren't resilient to different chunk sizes


TobiasPfeifer:
The `ZTransducer` for decoding fails when the input chunk doesn't exactly match one encoded object. Almost all rechunked streams will therefore fail. 
Transducers should be resilient to different chunk sizes (handle partial and excess input bytes).

```scala
// from ProtobufCodecSpec -> Should successfully encode and decode -> floats
val value = 0.001f
encodeAndDecode(Schema[Float], value) //<-- this effect will fail with "Invalid number of bytes for Float. Expected 4, got 2"

  def encodeAndDecode[A](schema: Schema[A], input: A): ZIO[Any, String, Chunk[A]] =
    ZStream
      .succeed(input)
      .transduce(ProtobufCodec.encoder(schema))
      .chunkN(2) // <-- test are fine without rechunking
      .transduce(ProtobufCodec.decoder(schema))
      .run(ZSink.collectAll)
```

petoalbert:
I'd like to work on this.

petoalbert:
I started looking into this issue but got stuck with the implementation. I see that zio-json already has a nice implementation for a [decoding transducer](https://github.com/zio/zio-json/blob/303e71cbcc4a3b0523919e7a77feaed1a9d6716b/zio-json/jvm/src/main/scala/zio/json/JsonDecoderPlatformSpecific.scala#L50). However, it only works on the JVM, because it uses a `java.io.Reader` implementation that blocks until new chunks arrive on the stream. I don't know yet how it could be done on Scala.Js without changing something in zio-json to fit this problem.

For protobuf, we could change the implementation from `Chunk[Byte] => (String, (Chunk[Byte], A)` to something like `ProtobufReader => IO[String, A]` whith `ProtobufReader` having a method `readBytes(count: Int): IO[E, Chunk[Byte]]`. Then we could provide a transducer implementation similar to zio-json but one that works with Scala.js as well. WDYT? Would you consider it an overkill to introduce ZIO effects to the decoder implementation just to solve this problem?

petoalbert:
Tokenizing the stream based on json syntax and then providing the buffered chunks to the decoder could also work, I can try that.

PR #245, opened 2022-05-04
Upgrade to ZIO 2.0.0-RC6

denisgarci:
This is not yet working, not sure how to handle the changes to `ZPipeline.utfDecode` in `JsonCodec.scala`.  The changes were made in this PR https://github.com/zio/zio/pull/6627.
I think I have to use a `ZSink.mkString` to convert the bytes into a Scala string but that would consume the stream. Is that the right approach?

### DIFF
--- a/project/BuildHelper.scala
+++ b/project/BuildHelper.scala
@@ -26,10 +26,10 @@ object BuildHelper {
   val Scala213: String   = versions("2.13")
   val ScalaDotty: String = "3.1.0" //versions.getOrElse("3.0", versions("3.1"))
 
-  val zioVersion        = "2.0.0-RC5"
-  val zioJsonVersion    = "0.3.0-RC7"
-  val zioPreludeVersion = "1.0.0-RC13"
-  val zioOpticsVersion  = "2.0.0-RC4"
+  val zioVersion        = "2.0.0-RC6"
+  val zioJsonVersion    = "0.3.0-RC8"
+  val zioPreludeVersion = "1.0.0-RC14"
+  val zioOpticsVersion  = "0.2.0-RC4"
   val silencerVersion   = "1.7.8"
 
   private val testDeps = Seq(
--- a/zio-schema-json/shared/src/main/scala/zio/schema/codec/JsonCodec.scala
+++ b/zio-schema-json/shared/src/main/scala/zio/schema/codec/JsonCodec.scala
@@ -23,7 +23,7 @@ object JsonCodec extends Codec {
     )
 
   override def decoder[A](schema: Schema[A]): ZPipeline[Any, String, Byte, A] =
-    ZPipeline.utfDecode >>> ZPipeline.mapZIO(
+    ZPipeline.fromChannel(ZPipeline.utfDecode.channel.mapError(_.toString)) >>> ZPipeline.mapZIO(
       (s: String) => ZIO.fromEither(Decoder.decode(schema, s))
     )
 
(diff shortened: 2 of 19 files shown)


adamgfraser:
The change is to the error type so you are going to need to transform the decoding error into a string so that it can be unified with the error type of the codec.

project/BuildHelper.scala
```diff
@@ -26,10 +26,10 @@ object BuildHelper {
...
   val ScalaDotty: String = "3.1.0" //versions.getOrElse("3.0", versions("3.1"))
 
-  val zioVersion        = "2.0.0-RC5"
-  val zioJsonVersion    = "0.3.0-RC7"
-  val zioPreludeVersion = "1.0.0-RC13"
-  val zioOpticsVersion  = "2.0.0-RC4"
+  val zioVersion        = "2.0.0-RC6"
+  val zioJsonVersion    = "0.3.0-RC8"
+  val zioPreludeVersion = "1.0.0-RC14"
+  val zioOpticsVersion  = "0.2.0-RC4"
```
denisgarci, line 32: As an aside, the previous release version for zio-optics probably got messed up, it was `2.0.0-RC4`.

denisgarci:
@adamgfraser Thank you will try that.

zio-schema-json/shared/src/main/scala/zio/schema/codec/JsonCodec.scala
```diff
@@ -23,7 +23,7 @@ object JsonCodec extends Codec {
     )
 
   override def decoder[A](schema: Schema[A]): ZPipeline[Any, String, Byte, A] =
-    ZPipeline.utfDecode >>> ZPipeline.mapZIO(
+    ZPipeline.fromChannel(ZPipeline.utfDecode.channel.mapError(_.toString)) >>> ZPipeline.mapZIO(
```
denisgarci, line 26: @adamgfraser I tried that, but it doesn't seem to work. I get `Unexpected end of input` in the `JsonCodecSpec` tests. Any suggestions?

vigoo: `utfDecode` tries to determine the encoding by reading a BOM from the beginning of the stream, and defaults to `UTF-8` if it could not find any.

I will check if there is something wrong with this logic but here you should be able to just use `utf8Decode` as JSON should always be in `UTF-8`.

vigoo: ..but that does not solve the issue, so I'm looking into it more

vigoo: I think you ran into this issue: https://github.com/zio/zio-schema/issues/182 because the new `utfDecode` rechunks the stream

denisgarci: @vigoo Thanks for your suggestions and thanks for adding `ZPipeline.mapError`. I still have the same error (`Unexpected end of input`) using `utf8Decode`. Will check https://github.com/zio/zio-schema/issues/182

bilal-fazlani:
👀

vigoo:
Let's release this for now until https://github.com/zio/zio-schema/issues/182 gets solved to unblock dependencies, as that will be a bigger change in both json and protobuf decoders.

vigoo (approved):

Issue #242, opened 2022-04-20
Capture the singleton type of the term of the record or enum and pass it through the accessor builder


sviezypan:
I need a singleton type of the term of the record or enum passed through AccessorBuilder, so something like:
```
trait AccessorBuilder {
  type Lens[F, S, A]
  type Prism[F, S, A]
  type Traversal[F, S, A]

  def makeLens[F, S, A](product: Schema.Record[S], term: Schema.Field[A]): Lens[F, S, A]

  def makePrism[F, S, A](sum: Schema.Enum[S], term: Schema.Case[A, S]): Prism[F, S, A]

  def makeTraversal[F, S, A](collection: Schema.Collection[S, A], element: Schema[A]): Traversal[F, S, A]
}
```

This need to modify also Accessors in Schema to 

```
type Accessors[Lens[_, _, _], Prism[_, _, _], Traversal[_, _, _]]
```

tusharmath:
If `S` is the whole object and `A` is a part of it, then what does `F` represent here?

sviezypan:
`F` would be a singleton type equal to the name of `A`, 

Maybe not useful in lot of places but lets say that we have 
`case class Person(name: String, age: Int)` 

and we need a type that represents tuple of its field names
`type FieldNames = ("name", "age")`

Then we can create an accessor builder like:
```
val identities = new AccessorBuilder {
    override type Lens[F, S, A] = F

    override type Prism[F, S, A] = Unit

    override type Traversal[S, A] = Unit

    def makeLens[F, S, A](product: Schema.Record[S], term: Schema.Field[A]): Lens[F, S, A] =
        term.label.asInstanceOf[F]

    def makePrism[F, S, A](sum: Schema.Enum[S], term: Schema.Case[A, S]): Prism[F, S, A] = ()

    def makeTraversal[S, A](collection: Schema.Collection[S, A], element: Schema[A]): Traversal[S, A] = ()
  }
  ```
And when we have personSchema we can call 
  ```
  type FieldNames = 
        personSchema.Accessors[
          identities.Lens,
          identities.Prism,
          identities.Traversal
        ]
 ```

sviezypan:
Implemented in #246

Issue #241, opened 2022-04-20
Capture the name of the record or enum inside the schema


sviezypan:
The same as Field[A] have a label, record and enums could have a name, so in the following example getName method would be able to return "Person" String with the help of Schema
```
case class Person(name: String, age:Int)
def getName[A](implicit i: Schema[A]) : String = ???

assertTrue(getName[Person] == "Person")
```

sviezypan:
Done in #246

PR #246, opened 2022-05-08
#241, #242 Capture name of product and enum to schema

sviezypan:
This PR implements features #241 and #242. However instead of adding `name` to all records and enums, `TypeId` was added so its possible to get fully qualified name like:

```scala
package zio.example {
  case class Person(age: Int)
  val personSchema = DeriveSchema.gen[Person]

  assertTrue(personSchema.id.fullyQualified == "zio.example.Person")
}
```

I am still missing macro schema generation for scala 3 which was recently added (scala 2 works fine)

In order to be able to convert from `Schema.Record` to `DynamicValue.Record` and back without loosing information `TypeId` was introduced also on `DynamicValue`. 
Because of that, however, I don't know how to write `Schema` for `DynamicValue.Record` as Schema[DynamicValue] describes DynamicValue itself, not underlying untyped data structure. So even though for example DynamicValue is representation of 
`Address(street: String, number:Int, postalCode: String)` 
its Schema is a CaseClass1
```scala
Schema.CaseClass1[Map[String, DynamicValue], DynamicValue.Record](
        TypeId.parse("zio.scheema.DynamicValue.Record"),
        Schema.Field("values", Schema.defer(Schema.map(Schema.primitive[String], DynamicValueSchema()))),
        map => DynamicValue.Record(TypeId.Structural, ListMap(map.toSeq: _*)),
        record => record.values
      )
```
As you can see Schema provides a way how to construct data structure, which in this case is DynamicValue, however we don't know the name of data structure that DynamicValue represents (Address) - we just know how to construct it. The same applies also for enums.

Files in `zio-schema-thrift/shared/src/test/scala-2/zio/schema/codec/generated/` were auto generated by new version of Thrift compiler 0.16.0 - no need to be reviewed.


jdegoes:
Thank you for your work on this! This is a highly sought-after addition to ZIO Schema.


>  So even though for example DynamicValue is representation of Address(street: String, number:Int, postalCode: String)
> its Schema is a CaseClass1

I think you have to use a hidden field inside the map, which stores the typeId for the record.

You can see it would affect this:

```scala
        map => DynamicValue.Record(/* extract out type ID from map */, ListMap(map.toSeq: _*)),
        record => record.values ++ /* store record type id into map */
```

The build is still failing, can you please take a look? 🙏

zio-schema/shared/src/main/scala/zio/schema/Schema.scala
```diff
@@ -393,11 +403,16 @@ object Schema extends SchemaEquality {
 
   final case class Tuple[A, B](left: Schema[A], right: Schema[B], annotations: Chunk[Any] = Chunk.empty)
       extends Schema[(A, B)] { self =>
-    override type Accessors[Lens[_, _], Prism[_, _], Traversal[_, _]] = (Lens[(A, B), A], Lens[(A, B), B])
+
+    val first = "_1"
+    val second = "_2"
```
sviezypan, line 408: I am using these strings as singleton type parameters. If I inline them it fails to compile on 2.12

sviezypan:
hey @jdegoes, took me some time to resolve issues for scala 2.12 and 3 but its ready to review now. Please check also the comment I made about extracting strings that I use for singleton types.
```
Prism[some.type, Option[A], Some[A]] // ok
Prism["Some", Option[A], Some[A]] // fails on 2.12
```

jdegoes (approved):

Issue #251, opened 2022-05-25
Add phone number validations for European countries


jdegoes:
**ISSUE**

There is no constructor for `zio.schema.validation.Validation` which can validate a European country phone number.

**NOTE**: This is 51 tickets in one. Please signup for one or more of the tickets by checking them below.

- [ ] Albania
- [ ] Andorra
- [ ] Armenia
- [ ] Austria
- [ ] Azerbaijan
- [ ] Belarus
- [ ] Belgium
- [ ] Bosnia and Herzegovina
- [ ] Bulgaria
- [ ] Croatia
- [ ] Cyprus
- [ ] Czechia
- [ ] Denmark
- [ ] Estonia
- [ ] Finland
- [ ] France
- [ ] Georgia
- [ ] Germany
- [ ] Greece
- [ ] Hungary
- [ ] Iceland
- [ ] Ireland
- [ ] Italy
- [ ] Kazakhstan
- [ ] Kosovo
- [ ] Latvia
- [ ] Liechtenstein
- [ ] Lithuania
- [ ] Luxembourg
- [ ] Malta
- [ ] Moldova
- [ ] Monaco
- [ ] Montenegro
- [ ] Netherlands
- [ ] North Macedonia
- [ ] Norway
- [ ] Poland
- [ ] Portugal
- [ ] Romania
- [ ] Russia
- [ ] San Marino
- [ ] Serbia
- [ ] Slovakia
- [ ] Slovenia
- [ ] Spain
- [ ] Sweden
- [ ] Switzerland
- [ ] Turkey
- [ ] Ukraine
- [ ] United Kingdom (UK)
- [ ] Vatican City (Holy See)

**SOLUTION**

Using other constructors which already exist in `zio.schema.validation.Validation` (which include regex validation, string length checking, and numeric range checking), introduce a new constructor that satisfies the requirements of this issue.

```scala
object Validation {
  ...
  val phoneNumber<2-Letter CountryCode>: Validation[String] = ???
}
```

**IMPACT**

Increasing the number of constructors for `Validation` will make it far easier to use the `validation` package to 

**WHERE TO LOOK**

 - [Schema validation](https://github.com/zio/zio-schema/pull/244)

**PROBLEMATIC ISSUES**

Issue #248 must be completed first, or at least, this issue must be based off of #244.

BalduinLandolt:
I'll try to get that one started by implementing Switzerland, but once I got the hang of it, I can continue with other countries too.

[...]

PR #255, opened 2022-05-26
Phone Number Validation Switzerland

BalduinLandolt:
First step towards #251: Switzerland


zio-schema/shared/src/main/scala/zio/schema/validation/Regex.scala
```diff
@@ -82,6 +83,19 @@ object Regex {
 
   def oneOf(chars: Char*): Regex = CharacterSet(chars.toSet)
 
+  def literalCharacter(char: Char): Regex = CharacterSet(Set(char))
+
+  def concatenation(str: String): Regex = {
```
jdegoes, line 88: I would call this `literal` because you are looking for a regular expression that matches a literal string.

Also please add in alphabetical order!

BalduinLandolt: done, changed it accordingly

jdegoes (approved):

jdegoes:
Thank you for your contribution!

PR #260, opened 2022-05-26
Fix protobuf semidynamic encoding/decoding

petoalbert:
Encoding of semidynamic values should conform to the protobuf encoding protocol. Each encoded object should be preceded by a length-limited tag. 

With the current setup, the semidynamic encoder just concatenates the encoded schema act bytes and the encoded value bytes. This causes issues in the following scenario:

Encode a top-level enum, whose case wraps a semidynamic schema. When decoding, the enum decoder will read the length-delimited tag for the encoded ast which doesn't include the total length (its missing the encoded value). This is because top-level values don't encode any tag, but the top-level enum decoder will still find a length-limited tag and skip the rest of the message. This is needed by zit-flow's Remote values, more details in this issue: https://github.com/zio/zio-flow/issues/144

My fix is that semidynamic should encode a tuple, so the attached length-delimited tag will include the correct message length.


zio-schema-protobuf/shared/src/main/scala/zio/schema/codec/ProtobufCodec.scala
```diff
@@ -565,8 +567,8 @@ object ProtobufCodec extends Codec {
           caseClass22Decoder(s)
         case Schema.Enum1(c, _)                                                                                                    => enumDecoder(c)
         case Schema.Enum2(c1, c2, _)                                                                                               => enumDecoder(c1, c2)
-        case Schema.Enum3(c1, c2, c3, _)                                                                                           => enumDecoder(c1, c2, c3)
         case Schema.Enum4(c1, c2, c3, c4, _)                                                                                       => enumDecoder(c1, c2, c3, c4)
+        case Schema.Enum3(c1, c2, c3, _)                                                                                           => enumDecoder(c1, c2, c3)
```
vigoo, line 571: Please undo this reordering :)

vigoo (changes requested):
Nice work, just a single comment

vigoo (approved):
Can you open another pull request against the zio 2 branch too?

vigoo (approved):

PR #262, opened 2022-05-26
Implement date-time validation

beem812:


jdegoes:
This approach is based on storing the format as a string. I don't know any protocols (including JSON Spec) that support this. So instead, I think we need to "compile" the format down to existing Validation constructors (almost certainly: Regex constructors). 

This means either parsing the format string, or getting Java to parse it, into some structured representation: and then translating the structured representation into the existing constructors (rather than adding a new constructor).

Does this make sense?

jdegoes:

Good to merge when the build is passing! Be sure to format your code using `scalafmt` commands in the console.

beem812:
Based on your first comment and the discussion in discord I'm wondering if the good to merge comment was correct? I know you're looking at deluge of PRs right now so just wanted to clarify. I assume that you'll be wanting to have an issue to come back and update this to use regex if you do mean to merge this.

jdegoes:
@beem812 Yes, if you can fix the conflicting files, I am happy to merge this in. Thanks for your work on this!

PR #265, opened 2022-05-27
Phone Number Validation Portugal

lmserrano:
Initial Phone Number Validation for #251 Portugal


domdorn:
I think we should extract the `PhoneNumberRegexs` into a own file, else the `Regexps` will soon have thousands of lines.

jdegoes (approved):

jdegoes:
@domdorn I agree! In fact, `PhoneNumberValidations` could be a separate trait in a separate file, mixed into the object.

domdorn:
@jdegoes when thinking about it, I realised that we have a similar concept for validation in zio-prelude: https://github.com/zio/zio-prelude/blob/master/core/shared/src/main/scala/zio/prelude/ZValidation.scala .. while this PR is probably the wrong scope, shouldn't we unify these concepts/datatypes and then create a distinct lib out of them, e.g. `zio-validation` (maybe even similar to JSR303 in functionality)?

lmserrano:
Thank you for the feedback @jdegoes . I've fixed some merge conflicts which had appeared after some more countries had been added by other people.

I've refactored into a separate file. Currently the object using the trait is its own though. Not sure how best to mix into Validation or other. Also am not familiar with zio-prelude and the issue raised by @domdorn , but feel free to edit and/or comment as necessary.

lmserrano:
I'm closing this Pull Request because I've only had the chance to look into this again now and while attempting to update it I noticed that the code in `main` branch already has some of the code from this PR adapted and added.

Issue #253, opened 2022-05-25
Add validation constructor to validate times


jdegoes:
**ISSUE**

There is no constructor for `zio.schema.validation.Validation` which can validate a local time in some specified format.

**SOLUTION**

Using other constructors which already exist in `zio.schema.validation.Validation` (which include regex validation, string length checking, and numeric range checking), introduce a new constructor that satisfies the requirements of this issue.

```scala
object Validation {
  ...
  def time(format: String): Validation[String] = ???
}
...
val validation: Validation[String] = Validation.time("hh:mm:ss")
```

**IMPACT**

Increasing the number of constructors for `Validation` will make it far easier to use the `validation` package to 

**WHERE TO LOOK**

 - [Schema validation](https://github.com/zio/zio-schema/pull/244)
 - [LocalTime](https://docs.oracle.com/javase/8/docs/api/java/time/LocalTime.html)
 - [DateTimeFormatter](https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html)

**PROBLEMATIC ISSUES**

Issue #248 must be completed first, or at least, this issue must be based off of #244.

tewecske:
I pick this one up for the hackathon.

tewecske:
Does it need to support the same stuff as the DateTimeFormatter?
```
   a       am-pm-of-day                text              PM
   h       clock-hour-of-am-pm (1-12)  number            12
   K       hour-of-am-pm (0-11)        number            0
   k       clock-hour-of-am-pm (1-24)  number            0

   H       hour-of-day (0-23)          number            0
   m       minute-of-hour              number            30
   s       second-of-minute            number            55
   S       fraction-of-second          fraction          978
   A       milli-of-day                number            1234
   n       nano-of-second              number            987654321
   N       nano-of-day                 number            1234000000
```

jdegoes:
@tewecske Do as much as you are willing. You may be able to just piggyback atop `DateTimeFormatter`, rather than implementing any new parsing. However, I didn't take a close look to see if they expose the (parsed) data.

tewecske:
Unfortunately the parsed data only available from `DateTimeFormatter.toString` as string. Like `Value(HourOfDay,2)':'Value(MinuteOfHour,2)':'Value(SecondOfMinute,2)`.

PR #272, opened 2022-05-29
Implemented #253 - Added time validation

tewecske:
Need suggestion if the approach is correct. @jdegoes suggested to use an ADT instead of the format string which might be a better approach. Right now I'm reimplementing DateTimeFormatBuilder but with less feature. DateTimeFormat(Builder) doesn't expose it's internals to get the parsed format from them.

The testing is also a bit special and I'm not sure if this is the proper way but I wanted to test a lot of things and this seemed an easy/fast way to validate parsing instead of manually creating all times. So I ended up comparing my validation to DateTimeFormat's parse for the same format and times. It doesn't completely match but I only left in the times which have the same parse result. For sanity I also check if I can correctly parse at least 25% of the example times so it can't happen that I think all is good but in reality neither my parser nor DTF could parse any of them. I'd like to add more formats but first I wanted to ask if this is okay?

I copied the symbol table javadoc and an error message from DTF and I don't know if that's "legal" :)

Also all other comments are welcome to help cleaning up the code :)

### DIFF
--- a/zio-schema/shared/src/main/scala/zio/schema/validation/Time.scala
+++ b/zio-schema/shared/src/main/scala/zio/schema/validation/Time.scala
@@ -1,0 +1,150 @@
+package zio.schema.validation
+
+import scala.annotation.tailrec
+import scala.collection.mutable
+
+trait Time {
+
+  /**
+   * Format is almost the same as the one used by the java.time.format.DateTimeFormatter class.
+   *
+   *  a           AM/PM always 2 letters
+   *  h           1-12 hour 1 or 2 digits
+   *  hh          01-12 hour always 2 digits
+   *  H           0-23 hour 1 or 2 digits
+   *  HH          00-23 hour always 2 digits
+   *  m           0-59 minute 1 or 2 digits
+   *  mm          00-59 minute always 2 digits
+   *  s           0-59 second 1 or 2 digits
+   *  ss          00-59 second always 2 digits
+   *
+   *  S           0-9 fraction of seconds 1 digits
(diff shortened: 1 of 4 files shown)


zio-schema/shared/src/main/scala/zio/schema/validation/Time.scala
```diff
@@ -0,0 +1,139 @@
...
+
+import scala.annotation.tailrec
+import scala.collection.mutable
+
+trait Time {
+
+  /**
+   *
+   * Format is the same is the one used by the java.time.format.DateTimeFormatter class.
+   * This is copied from there.
```
jdegoes, line 11: Unfortunately we can't copy even comments from the JDK without running into issues. I'd suggest linking to the Javadoc or re-creating a similar table that documents the format accepted.

jdegoes (changes requested):
One minor comment on the Javadoc but otherwise looks good. Thanks for your work on this!

jdegoes:
Ah, be sure to resolve the conflicts too, so the CI will build!

tewecske:
I removed the comments and the error message which were copied from JDK and created a new ones.

zio-schema/shared/src/main/scala/zio/schema/validation/Time.scala
tewecske: Changed it.

jdegoes (approved):

jdegoes:
Thank you, and congrats on your PR!

Issue #274, opened 2022-05-31
Ensure patches (Diff) can be inverted to take the new value back to the old value


jdegoes:

robmwalsh:
I'll take this one

PR #276, opened 2022-06-05
Inverse #274

robmwalsh:
Some (very) initial work on #274

A fundamental problem to solving this is that not all diffs currently can be inverted, which mainly comes down to what I'd consider "failure cases" in diffs (especially NotComparable). I don't think there's any reason a diff between two objects of the same type should ever fail, or should ever not be invertable, so anything that can't satisfy the law proposed here (https://github.com/zio/zio-schema/issues/274#issuecomment-1143338021) needs to be reworked.

Patching can certainly fail (though it should never fail between the two objects that generated the diff), but I think it would make more sense to return these failures in a a purpose built data structure which captures the partially applied diff and can be used to resolve conflicts, rather than strings

~~The concept of a (3 way) merge also seems to be missing~~ this requires history and a common ancestor which isn't known here. Perhaps one for later.

@jdegoes any thoughts on this?


jdegoes:
I agree we need to focus on creating a "diff algebra" that is fully invertible. If we need a concept of partial invertibility, we should be very careful in limiting its scope.

robmwalsh:
@jdegoes I've done a fair bit of digging about, and I think that something based on this would be the best option:
https://www.merlin.uzh.ch/publication/show/2531
https://bitbucket.org/sealuzh/tools-changedistiller/src/master/

This is aimed at ASTs but should easily be adaptable to any tree structure.
It's... not small! but I think it's going to be the best solution because it can identify add, change, move and delete operations. This could be done on `DynamicValue` initially, which would support all cases. In future, deriving a differ based on a Schema would probably be a useful exercise. Other tree diffing algorithms are more limited in scope, supporting only add, delete and relabel (which isn't valid for this use case). See section 6 for a comparison with these other algorithms (unfortunately I found and waded through all of these other algorithms first!).

jdegoes:
This has bit-rotted, so will close for now. The concept of inverse patches needs additional development in main, so feel free to re-open if you get more bandwidth to work on this!

Issue #249, opened 2022-05-25
Add phone number validations for strings representing US phone numbers


jdegoes:
**ISSUE**

There is no constructor for `zio.schema.validation.Validation` which can validate a US phone number.

**SOLUTION**

Using other constructors which already exist in `zio.schema.validation.Validation` (which include regex validation, string length checking, and numeric range checking), introduce a new constructor that satisfies the requirements of this issue.

```scala
object Validation {
  ...
  val phoneNumberUS: Validation[String] = ???
}
```

**IMPACT**

Increasing the number of constructors for `Validation` will make it far easier to use the `validation` package to 

**WHERE TO LOOK**

 - [Schema validation](https://github.com/zio/zio-schema/pull/244)

**PROBLEMATIC ISSUES**

Issue #248 must be completed first, or at least, this issue must be based off of #244.

jadireddi:
Would like to contribute.

PR #277, opened 2022-06-05
Phone Number Validation

uurl:
This PR solves #249, solves #250 and solves #251 with all the countries in the world


zio-schema/shared/src/main/scala/zio/schema/validation/PhoneNumberValidation.scala
```diff
@@ -0,0 +1,2383 @@
...
+  lazy val phoneNumberGH: Validation[String] = {
+    val countryCode = Regex.literal("233")
+    val prefix      = internationalPrefix ~ optionalSeparator ~ countryCode | nationalPrefixZero
+    val phoneNumber = digitsWithSeparator.between(9, 10)
+
+    Validation.regex(prefix ~ optionalSeparator ~ phoneNumber)
+  }
+
+  /** Phone number validation for Gibraltar */
+  lazy val phoneNumberGI: Validation[String] = {
```
vigoo, line 805: I see that this is a very common pattern, probably would worth moving into a method with parameters for the country code and maybe the min/max number of digits to reduce code duplication.

(or maybe a couple of such methods if there are other common cases shared by many countries)

The more unique cases could remain as they are, of course.

uurl: Good question, the point here is that each specific country has its own dialing rules and its own prefixes, in fact, the original issue only asked for US, UK and Europe. For example, as it stands, we can add whatever rules we want to the US and UK individually, but I don't really know what they want the scope of this validation to be.  I'm going to refactor the code, but adding new specific validations for each country would lead to the same problem.

jdegoes (approved):

jdegoes:
What a massive undertaking! Would be nice at some point to factor out some common code, but I am so impressed by the scope of this PR and grateful for this contribution, I am happy to merge as-is. Thank you so much & congratulations on your PR!

subotic:
@uurl please don't just take other people's code without proper attribution. Not a nice thing to do.

uurl:
A proper attribution of 0.27% (10/3722) of this PR to @subotic , without his effort it wouldn't have been possible

jdegoes:
@uurl Proper attribution is extremely important in the ZIO community, and should either happen by officially basing your work off another person's pull request, so that they will receive "Github credit" for it, or in cases where this is not possible, crediting the developer with a note in the source code.

Thanks for your work building on @subotic's pull request, and to ensure he receives proper credit, please submit a followup pull request that, in a comment, notes the lines of code that should be credited.

Something like:

```scala
// Thanks to @subotic for his work developing this code
```

It might also be possible to git rewrite history, but I can't help you with that.

@subotic Thanks for writing code that formed part of this pull request, and we'll make sure you receive full attribution for it!

PR #287, opened 2022-06-18
In JsonCodec: skip fields with null values in caseClassEncoder (make it consistent with zio-json)

pwliwanow:
Thanks for this wonderful library!

This PR makes writing fields with null values for case classes consistent with zio-json.

For example, given the following code:
```scala
final case class Person(age: Option[Int], name: String)
val person = Person(None, "John")
```

existing code will encode the `person` instance as: `{"age": null, name: "John"}`, while zio-json would encode it as: `{"name":"John"}`. Code in the PR makes `JsonCodec` consistent with zio-json in this case.


jdegoes (approved):

jdegoes:
Thank you!

Issue #258, opened 2022-05-26
Add a validation constructor for IPv6


digy:
Implement an ipv6 validator

baldram:
I will take this one.

baldram:
Hey, I was working on the IPv6 validation and realised it's non-trivial. 
It's because of the different notations. There are many nooks and crannies of rules to explore.
The implementation delivered by @uurl is the first good step for further extension. 

What's missing?
Example:
✅  The "compressed notation" works ok: `1762:0:0:0:0:B03:1:AF18` -> `1762::B03:1:AF18`.
🟥  The "compressed mixed notation" does not: `1762:0:0:0:0:B03:127.32.67.15` -> `1762::B03:127.32.67.15`, as only the first one works and the compressed form is rejected.

I will try to prepare more test cases to find various border cases.

There is also one thing to clarify. A test case in the current implementation expects a valid IPv6 address to be incorrect. I mean this test scenario: `2001:cdba:0000:0000:0000:0000:3257:9652`. It seems to be a correct IPv6 address.
Would you please clarify @uurl? According to my research, I would fix it and move this scenario to valid ones.

I prepared the ground for extensions and fixes via some other PRs (merged), and I will try to continue here if you don't mind.

uurl:
> Hey, I was working on the IPv6 validation and realised it's non-trivial. It's because of the different notations. There are many nooks and crannies of rules to explore. The implementation delivered by @uurl is the first good step for further extension.
> 
> What's missing? Example: white_check_mark The "compressed notation" works ok: `1762:0:0:0:0:B03:1:AF18` -> `1762::B03:1:AF18`. red_square The "compressed mixed notation" does not: `1762:0:0:0:0:B03:127.32.67.15` -> `1762::B03:127.32.67.15`, as only the first one works and the compressed form is rejected.
> 
> I will try to prepare more test cases to find various border cases.
> 
> There is also one thing to clarify. A test case in the current implementation expects a valid IPv6 address to be incorrect. I mean this test scenario: `2001:cdba:0000:0000:0000:0000:3257:9652`. It seems to be a correct IPv6 address. (Leading zeros are optional and it does not have to be `2001:cdba:0:0:0:0:3257:9652`). Would you please clarify @uurl? According to my research, I would fix it and move this scenario to valid ones.
> 
> I prepared the ground for extensions and fixes via some other PRs (merged), and I will try to continue here if you don't mind.

[...]

PR #301, opened 2022-07-26
Feature: Support IP in email address validation

baldram:
The domain part of the email may be a domain name or an IP address enclosed in brackets. See: https://en.wikipedia.org/wiki/Email_address

Let's use the power of the composition:
- use IPv4 validation as a subset of IPv6
- use IP validation from email validation
- bring a missing feature to email validation
- introduce parameterized tests for readability 

This PR is related to issue https://github.com/zio/zio-schema/issues/258


jdegoes (dismissed):

baldram:
Hey @jdegoes thanks for approval. A change was approved but not yet merged, so I added here my next PR as it's related.

All stuff is done.
Would you please have a look again if we're ready to go?

jdegoes (approved):

jdegoes:
🙏

Issue #248, opened 2022-05-25
Get #244 merged


jdegoes:
**ISSUE**

A lot of additional tickets depend on the successful merging of #244, which enhances ZIO Schema with a new package, `zio.schema.validation`, which has very simple support for defining validations. Though this work is not yet complete, it is useful in its own right, and merging it will enable others to build off and extend the work.

**SOLUTION**

To solve this issue, clone the repository and grab the branch, fix all compilation errors, format the code, delete any printlns or other code not suitable for production (examples in prodution code), and ensure all tests are passing, before submitting a separate pull request that is fully ready to merge. No additional features need be added to get the work in suitable shape for merging.

**IMPACT**

This will add first-class validation support into ZIO Schema and open the way for many additional tickets.

**WHERE TO LOOK**

 - [Pull Request #244](https://github.com/zio/zio-schema/pull/244)
 - [The repo to clone / fork](https://github.com/AppleKid7/zio-schema/tree/schema-validation)

**PROBLEMATIC ISSUES**

None.

digy:
Will work on this.

digy:
Created a PR https://github.com/zio/zio-schema/pull/254

jdegoes:
Fixed in #254!

PR #303, opened 2022-08-10
#248: Respect RFC 3696 for email length limits

baldram:
Related to #248 

This change improves email address validation and supports the [RFC 3696](https://www.rfc-editor.org/errata_search.php?rfc=3696&eid=1690) (including "[erratum 1003](https://www.rfc-editor.org/errata_search.php?rfc=3696&eid=1690)").

There is a length limit on email addresses. That limit is a maximum of 64 characters (octets) in the "local part" (before the "@") and the upper limit on address lengths - total size of 254 characters due to a restriction in RFC 2821.

Additionally:
- refactored test cases and introduced parametrized tests for clarity,
- added new test examples to cover this change,
- added some bonus tests to cover a few border cases that were luckily supported but not documented,
- a validation still has open points for separate PR (e.g. IP addresses support instead of domain - currently blocked by not-reviewed/unmerged PR #301; see also IP email examples: https://en.wikipedia.org/wiki/Email_address#Examples)

### DIFF
--- a/zio-schema/shared/src/main/scala/zio/schema/validation/Regexs.scala
+++ b/zio-schema/shared/src/main/scala/zio/schema/validation/Regexs.scala
@@ -5,20 +5,17 @@ trait Regexs {
   val identifier: Validation[String] =
     Validation.regex((Regex.digitOrLetter | Regex.oneOf('_')).atLeast(1))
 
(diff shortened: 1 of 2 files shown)


jdegoes (approved):

jdegoes:
🙏

PR #318, opened 2022-09-05
Refactoring:  introducing Patch data type

### LINKED ISSUE #317: Refactoring: introducing Patch data type
Patch data type will replace the `Diff[A]` that it is a better name and patch should be invertible.

devsprint:
Fixes #317


devsprint:
another implementation done.

PR #321, opened 2022-09-06
Refactoring introducing typetag

devsprint:


devsprint:
@jdegoes can you take a look to last commit where I have added `zio.Tag` ? The issue is that I have to propagate it everywhere and I think that it is not correct.

PR #326, opened 2022-09-09
Refactoring: renames

devsprint:
Renamed:
SetSchema into Set
Tuple in Tuple2
MapSchema into Map
EitherSchema into Either


devsprint:
already merged.

PR #329, opened 2022-09-23
Add methods for getting JsonEncoder/Decoder directly

kitlangton:
`JsonCodec.encode` currently generates a new `JsonDecoder` every time an `A` is decoded. For `zio-http` (and other libraries) it should be more efficient to precompute the `JsonCodec` and then call it directly.


vigoo (approved):
I think all codec implementations suffer from this and it would be nice to require a method in the `Codec` trait itself that "fixes" the codec to a given type (and returns a `FixedCodec[A]` or something

Issue #330, opened 2022-09-23
Cannot serialize DynamicValue.SetValue 


vigoo:
Fails with
- JSON
- Protobuf
- Thrift

(and there are no tests for dynamic value serialization for Avro at all)

mjabari:
I fix it

vigoo:
> I fix it

Thank you!

PR #331, opened 2022-09-23
Tests for serializing dynamic sequence and set

vigoo:
Reproducer for #330 

TODO: fix


jdegoes:
@vigoo I feel we should get this merged in, with the tests marked "@@ ignore", and referenced in issues. What do you think?

jdegoes (approved):

PR #335, opened 2022-09-26
Typeid changes

robmwalsh:
WIP adding some type safety to schemas because there are too many strings!

Mostly working but needs tidy up particularly around limiting where/how `TypeId`s are created
Scala 3 macro implementation hasn't been touched at this stage.

I'm currently stuck with fixing the schema derivation - the types are mapped to concrete types which mean `TypeId` and `FieldId` both become `String`s, which fails because the Schema cases now require either `TypeId` or `FieldId` (both prelude `Subtype[String]`s)

Any suggestions on how to do this?


zio-schema-derivation/shared/src/main/scala-2/zio/schema/DeriveSchema.scala
```diff
@@ -179,15 +179,17 @@ object DeriveSchema {
         val arity = fieldTypes.size
 
         val typeArgs = fieldTypes.map { ft =>
-          val fieldType = concreteType(tpe, tpe.decl(ft.name).typeSignature)
-          q"$fieldType"
+          val fieldType = concreteType(tpe, tpe.decl(ft.name).typeSignature) //need to not map TypeId/FieldId to String
+          if (fieldType <:< c.typeOf[FieldId]) //this doesn't work
```
robmwalsh, line 183: The problem!

jdegoes:
This has heavily bit-rotted. Please re-open if you get a chance to continue development!

Issue #333, opened 2022-09-25
Schema and DynamicValue should preserve tuple structures when schema got serialized 


vigoo:
The simple round-trip test for DynamicValue:

Given a value `a: A` and a `schema: Schema[A]` if we convert it to dynamic value and back: 

```scala
schema.fromDynamic(schema.toDynamic(a)))
```

we get back the same value `a`. This property is verified by tests and it is true for all the supported cases.

However if we serialize the schema and read it back:

scala
```
val schemaAst = schema.ast
val schema2   = schemaAst.toSchema
```
and try to materialize the dynamic value with `schema2`, simulating that the value and its schema was sent to another process,
the property is no longer true for all cases:

- Records and enums are materialized as generic `ListMap[String, _]` values which is expected
- However we also loose the structure of tuples! 
  - higher arity tuples are converted back to a nested representation `(A, B, C)` becomes `((A, B), C)`

I think for tuples we should be able to reconstruct the original tuple arity in this case.

mjabari:
I'll work on this one

jdegoes:
[Here is a reproducer](https://github.com/zio/zio-schema/pull/334/files)

PR #355, opened 2022-10-10
fix for Issue 333 WIP

mjabari:
#333
I examined many ways to fix this. despite all limits (no TypeTag, no Transform and ...) I think it's the cleanest. 
another drawback is that MetaSchema.Tuple uses CaseClass4 so it occupies 2 more slots
a hint would be welcomed


jdegoes:
@mjabari Thanks for your work on this!

@vigoo Is this the direction you had in mind?

vigoo:
@mjabari thanks! 
@jdegoes I did not have any particular solution in mind. this would most likely work, although it brings 22 new constructors to one more type which makes it harder to work with that; if I understand correctly, it is enough to do it in `MetaSchema` and `DynamicValue` is would still have only a `Tuple2`? 

About the limitation of "2 more slots" I don't think that's a blocker because you can also introduce nesting (separate metadata and per-field schemas) or use a list of `MetaSchema`s.

Actually wouldn't it be possible to be a bit more dynamic here, and only have a single `MetaSchema.Tuple` with a list of child schemas? Just to avoid the 21 new constructors. Just an idea

jdegoes:
@mjabari There are a bunch of conflicts. Would you please re-open this PR if you get a chance to fix them? Thank you!

PR #363, opened 2022-10-11
refactor: Scala 3 increase stack size limit

tusharmath:


zio-schema-derivation/shared/src/main/scala-3/zio/schema/macros.scala
```diff
@@ -38,8 +38,7 @@ private case class DeriveSchema()(using val ctx: Quotes) extends ReflectionUtils
 
   def deriveSchema[T: Type](stack: Stack = Stack.empty, top: Boolean = false): Expr[Schema[T]] = {
     depth += 1
-    // println(s"STACK ${stack} ${TypeRepr.of[T].show}")
-    if (depth > 85)
```
tusharmath, line 42: Not sure why this is `85`, I have updated it to a value that works for me :P

jdegoes (approved):

PR #366, opened 2022-10-11
Schema validation

AppleKid7:
* fixed Scala 3 macros for validations
* added validation method to Schema class
* added several Schema validation tests for both Scala 2 and Scala 3


AppleKid7:
I'd like to mention a couple of things: 

* the linter was giving me errors when I did not add the type ascriptions for some of the schema derivations, which I know isn't ideal. Is there a way around this? If so, I'd happily make the change.
* the validation function used to have cases for `SemiDynamic` and `MetaSchema`, but those two seem to be gone as case classes after a recent refactoring, so I removed them because I think there's no point in pattern-matching on them (plus it was giving me errors). Can anyone confirm this is ok?

jdegoes:
@AppleKid7 Thank you! Can you fix conflicts?

jdegoes:
@AppleKid7 I will close this PR temporarily. Please re-open when you have a chance to fix merge conflicts. Thank you!

AppleKid7:
Will do. I'll have the conflicts resolved within the next couple of days!

Issue #164, opened 2021-12-14
Schema isomorphism


jdegoes:
Increasingly, it will be necessary for _ZIO Schema_ to support schema-first approaches: for example, a new Scala project must interoperate with existing gRPC services, whose messages are defined using protobuf. 

In such case, the Scala-first approach supported by _ZIO Schema_ will not work very well, because it would require Scala developers to painstakingly create an ADT whose automatically-derived Schema will "happen" to correspond with the schema-first schema.

A much better approach is for code-generators (or ideally, macros) to programmatically create a ZIO Schema. This ZIO Schema would ideally act as a canonical definition of the protocol, but without necessarily defining or constraining what user-defined ADT a Scala developer could use to interact with that protocol.

Thanks to migrations, this approach is already possible: code-gen could generate a ZIO Schema, and then a user could "migrate" that Schema to another one, for a user-defined ADT. The drawback to this approach is that the migration cannot known to be valid or invalid at compile-time: one must actually try it at runtime to see if the migration is possible.

A way of pulling this validation to compile-time may be introducing a notion of _schema isomorphism_. Independently, this concept has arisen in _ZIO SQL_, which @sviezypan is working on. In particular, Jaro created a type class to witness that a given schema had all the right types in all the right orders to allow inserting case classes into a relational database table.

It may be the case that pulling this concept of compile-time "schema isomorphism" may enable schema-first usage of ZIO Schema, without sacrificing all the type safety and value-oriented features that ZIO Schema provides, while enabling richer integration in libraries like ZIO SQL, without them having to introduce their own more limited notions.

How might such a change be accomplished? 

One idea is adding another type parameter to `Schema`, such that `Schema[A]` would become `Schema[Abstract, Concrete]`. The new type parameter would represent the structure of the data type, independent of ordering or specific materialization.

For example (assuming Scala 2.13 and higher, with singleton types):

```scala
type ->*[A, B]
type ->+[A, B]
// type &[A, B] = A with B

final case class Person(name: String, age: Int)

implicit val schema: Schema[("name" ->* String) & ("age" ->* Int), Person] = ...
```

With such a type parameter, which is not modified by `Transform` nodes, it becomes possible to do type-level comparisons and transformations on the structure of a schema. Moreover, it becomes possible in libraries like ZIO SQL to say, _"I can work with any schema, so long as it has N fields of types X, Y, Z, ..."_.

[...]

PR #373, opened 2022-10-18
#164 implemented schema isomorphism for Products

sviezypan:
This PR adds Terms and FieldNames types to all product schemas. 
Terms consist of intersection types of tuples of field names and types while FieldNames is intersection type of field names.
Example for `CaseClass3`
```
    override type Terms = (F1, A1) with (F2, A2) with (F3, A3)

    override type FieldNames = F1 with F2 with F3
```
Where `F1` represents singleton type of field name and `A1` is field type.

Signature of `Field` is changed from `Field[A]` to `Field[F <: Singleton with String, A]` and its label is not of type `String` but `F`.

For this to work I had to introduce breaking changes for CaseClassX - they contain now all the F type parameters so 
CaseClass3 looks like the following:
```
sealed case class CaseClass3[F1 <: Singleton with String, F2 <: Singleton with String, F3 <: Singleton with String, A1, A2, A3, Z]
```


jdegoes:
Excellent work!

I wonder if, before merging this, we should see if we can push the type parameters as type members?

For example:

```scala
final case class CaseClass2[A, B, Z](...) {
  type FieldName1
  type FieldName2
  final type Terms = (FieldName1, A) with (FieldName2, B)
}
```

Then if the user is constructing such a thing manually (and painfully), they won't bother to specify the field name types (they'll be unknown and not usable for schema isomorphisms). It's only the macro that will generate this information appropriately.

What do you think?

sviezypan:
@jdegoes At first sight, it seems possible, however Macro would have to return CaseClass3.Aux instead of CaseClass3.Aux
```
  object CaseClass3 {
    type Aux[F1 <: Singleton with String, F2 <: Singleton with String, F3 <: Singleton with String, A1, A2, A3, Z] = 
      CaseClass23[A1, A2, A3, Z] {
        type FieldName1 = F1
        type FieldName2 = F2
        type FieldName3 = F3
      }
  }
```

As you said, manually derived case class would not have meaningful FieldNameX, I have to check what implications it would have on makeAccessors(), where Lens has access to FieldName.
```
type Lens[FieldName <: Singleton with String, S, A]
```

The only problem is with `FIeld` which is in the constructor of CaseClass, and which takes also FieldName as a type parameter.
```
sealed case class CaseClass1[FieldName <: Singleton with String, A, Z](
    id: TypeId,
    field: Field[FieldName, A],
    ...
  )
```

Would you change FIeld's label
```
 final case class Field[FieldName <: Singleton with String, A](
    label: FieldName,
    ...
  )
```
back to String and throw away `FieldName` type parameter from `Field`?
```
```
 final case class Field[A](
    label: String,
    ...
  )
```
```

jdegoes:
@sviezypan We could have `Field.Aux` if necessary. What do you think?

sviezypan:
@jdegoes I just don't see how Field.Aux is gonna help us. `Field` is inside the constructor of `CaseClassX`. If we swap `Field[A]` with `Field.Aux[FieldName, A]` we still need that `FieldName` type parameter declared somewhere.

Otherwise, drawback of this approach is additional type parameters, drawback of the other approach is the return type of `Aux` from the generated schema. I am not sure which one is better/worse.

jdegoes:
> I just don't see how Field.Aux is gonna help us. Field is inside the constructor of CaseClassX. If we swap Field[A] with Field.Aux[FieldName, A] we still need that FieldName type parameter declared somewhere.

Inside `CaseClassX`, we would have them as type member already, as `FieldName1`, etc.

Something like:

```scala
trait CaseClass2[A, B, Z]( {
  type FieldName1
  type FieldName2

  val field2: Field.Aux[FieldName1, A]
  val field2: Field.Aux[FieldName2, B]
}
```

Right?

sviezypan:
@jdegoes are you suggesting to change Field and all 22 overrides of CaseClassX from case class to trait? 

Each one of those would have a companion object with Aux type (for return type of macro) and abstract fields and abstract type members would still need to be specified when constructing this trait manually, right? 
So I am not sure if its beneficial over CaseClass with type parameters - those could be omitted when pattern matching against case class.

sviezypan:
We'll try different approach with Fields as type members

PR #389, opened 2022-10-27
Make DynamicValue.fromSchemaAndValue and codecs stack safe

### LINKED ISSUE #286: Recursive data-type can't be encoded with `ProtobufCodec`
A schema for the following recursive data-type derives without any errors, but then the test blows up (stack overflow) at runtime during encoding by `ProtobufCodec`:

```scala
object EncodingSpec extends DefaultRunnableSpec {

  sealed trait Recursive
  case class Case1(r: Either[Recursive, String]) extends Recursive

  val s: Schema[Recursive] = DeriveSchema.gen

  val spec = suite("CodecsSpec")(
    test("encoding") {
      val r = Case1(Left(Case1(Right("OK"))))
      val bytes = ProtobufCodec.encode(s)(r)
      assert(ProtobufCodec.decode(s)(bytes))(isRight(equalTo(r)))
    },
  )
}
```

Tested with zio-schema v0.1.9.

### LINKED ISSUE #340: DynamicValue.fromSchemaAndValue is not stack safe
For large data structures such as `zio-flow`s `ZFlow` it quickly runs out of available stack. It would be nice to have a stack-safe implementation.

vigoo:
Resolves #340 
Resolves #286


vigoo:
Work in progress - making a generic way to deal with stack safety and updating other parts of the library to build on that

adamgfraser (dismissed):

adamgfraser (dismissed):

PR #390, opened 2022-10-27
Fixed #376 - Implement support for @discriminatorName

### LINKED ISSUE #376: Implement support for @discriminatorName

This ticket is to implement the `@discriminatorName` annotation. This annotation, which may be applied to any `sealed trait` (enumeration), means to specify the name of a field that will be used to contain the identity of the case of the enum that is being serialized.

For example, it is common in JSON to use a field named `type` to encode which case of an enum (such as `Either`) is being stored (e.g. `"typeId": "Right"` versus `"type": "Right"`.

_ZIO Schema_ already captures annotations on records and enumerations. So the main work of this ticket will be to look for the annotation `zio.schema.annotation.discriminatorName` on any sealed trait (enumeration), and then leverage it in any of the existing codecs for which it makes sense.

 - [ ] JSON Codec

exp0nge:
* adds new annotation for discriminatorName
* replaces type name for the discriminator key-value pair


jdegoes:
Can you fix the [2.13 compile-time failure](https://github.com/zio/zio-schema/actions/runs/3341197601/jobs/5542320951#step:5:500)?

exp0nge:
> Can you fix the [2.13 compile-time failure](https://github.com/zio/zio-schema/actions/runs/3341197601/jobs/5542320951#step:5:500)?

Done, do these run nightly?

zio-schema-json/shared/src/test/scala-2/zio/schema/codec/JsonCodecSpec.scala
```diff
@@ -660,11 +670,15 @@ object JsonCodecSpec extends ZIOSpecDefault {
...
 
-  private def assertEncodes[A](schema: Schema[A], value: A, chunk: Chunk[Byte]) = {
+  private def assertEncodes[A](schema: Schema[A], value: A, chunk: Chunk[Byte], log: Boolean = false) = {
     val stream = ZStream
       .succeed(value)
       .via(JsonCodec.encoder(schema))
       .runCollect
+      .tap { chunk =>
+        if (log) printLine(s"${new String(chunk.toArray)}").ignore
+        else print("").ignore
```
exp0nge, line 680: not sure how to do a no-op here

jdegoes: `ZIO.unit`

zio-schema-json/shared/src/test/scala-2/zio/schema/codec/JsonCodecSpec.scala
```diff
@@ -140,6 +141,15 @@ object JsonCodecSpec extends ZIOSpecDefault {
           Enumeration(StringValue("foo")),
           JsonCodec.Encoder.charSequenceToByteChunk("""{"oneOf":{"StringValue":{"value":"foo"}}}""")
         )
+      },
+      test("ADT with annotation") {
```
jdegoes, line 145: We need a test for decoding, too! To ensure we can round-trip: go from our case class to JSON, and go back from our JSON to the same case class.

zio-schema-json/shared/src/main/scala/zio/schema/codec/JsonCodec.scala
```diff
@@ -222,20 +224,35 @@ object JsonCodec extends Codec {
...
       (value: Z, indent: Option[Int], out: Write) => {
         val fieldIndex = cases.indexWhere(c => c.deconstructOption(value).isDefined)
         if (fieldIndex > -1) {
           val case_ = cases(fieldIndex)
-          out.write('{')
+          val discriminatorChunk = parentSchema.annotations.collect {
+            case d: discriminatorName =>
+              parentSchema.id match {
+                case TypeId.Structural => None
+                case n: TypeId.Nominal => Some((d, (n.packageName ++ n.objectNames ++ Chunk(case_.id)).mkString(".")))
```
jdegoes, line 236: Rather than using a type ID, we should use the `id` of the `Case` of the enumeration that we are encoding. That is used even when discriminators are not used to identify which of the case in the enum a value is.

jdegoes (changes requested):
Just two minor comments, and then will be good to merge. Thanks for your work on this important feature!

exp0nge:
> Just two minor comments, and then will be good to merge. Thanks for your work on this important feature!

comments addressed & commits squashed - f5b5caa89fe7138ad634df24dea7c80b2c4f8d46

jdegoes:
@exp0nge There are a few conflicts, I am happy to merge this as soon as these conflicts are resolved. Thanks again for your work on this!

exp0nge:
> @exp0nge There are a few conflicts, I am happy to merge this as soon as these conflicts are resolved. Thanks again for your work on this!

Done!

jdegoes:
@exp0nge Hate to ask one more time, but there are a few small conflicts! If you can get these fixed by tomorrow, I can merge your PR before anything else!

exp0nge:
> @exp0nge Hate to ask one more time, but there are a few small conflicts! If you can get these fixed by tomorrow, I can merge your PR before anything else!

✅

jdegoes (approved):

jdegoes:
Merged! Thanks so much for your patience. 🙏

PR #396, opened 2022-10-28
Implements support for @fieldName

### LINKED ISSUE #377: Implement support for @fieldName

This ticket is to implement the `@fieldName` annotation. This annotation, which may be applied to any field of a `case class` (record), means to specify the "serialization" name of the field to which it is applied.

For example, it is common in relational databases to name columns with snake case (`user_id`) rather than camel case (`userId`). Yet, in our Scala data types, we may wish to use camel case. So applying `@fieldName("user_id")` to a field `userId: Int` would be an instruction to use the alternate name rather than the Scala name.

The main place where this new annotation should be used is in Schema auto-derivation. Right now, ZIO Schema will generate `Schema.Field` value for each field inside a case class. These fields use the Scala field names. But, in the presence of this annotation, they should instead use the specified name.

There are two ways to implement this ticket:

1. Modifying the Scala 2 / Scala 3 macros to generate a different value for the field names, if this annotation is present.
2. Post-processing the auto-generated Schema to incorporate changes brought about by annotations such as `fieldName`.

juliano:
Fixes #377

### DIFF
--- a/zio-schema-derivation/shared/src/main/scala-2/zio/schema/macros.scala
+++ b/zio-schema-derivation/shared/src/main/scala-2/zio/schema/macros.scala
@@ -160,6 +160,11 @@ object DeriveSchema {
           }
         }
 
+    def getFieldName(annotations: List[Tree]): Option[String] =
+      annotations.collectFirst {
+        case q"new fieldName($name)" => name.toString
+      }.map(s => s.substring(1, s.length() - 1))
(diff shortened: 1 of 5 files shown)


jdegoes (approved):

jdegoes:
An incredibly useful addition to ZIO Schema. Thank you!

PR #405, opened 2022-10-31
added scala 2.x macro derivation for Map

### LINKED ISSUE #365: `DeriveSchema.gen[scala.collection.Map[K, V]]` does not work
Not sure if this would be considered a new feature or a bugfix.

Attempting to derive a schema from a Map[K,V] results on the following error:

```
Failed to derive schema for scala.collection.Map[String,zio.schema.SchemaValidationSpec.Person]. Can only derive Schema for case class or sealed trait
[error]       implicit val schema = DeriveSchema.gen[scala.collection.Map[String, Person]]
```

devsprint:
Fixes #365


jdegoes (approved):
It's' possible to fix this in the macro (as you have done), but I wonder if the problem is the implicit schema for map is not being found by implicit summoning in the macro?

PR #411, opened 2022-11-03
Error hierarchy

### LINKED ISSUE #387: Introduce proper error hierarchy for codecs
Currently, decoding methods in `zio.schema.codec` use a `String` for an error type, which is not sufficient to capture structured details, nor is it easily compatible with `Exception` / `Throwable`, common error types in Scala / Java.

To fix these issues, we can introduce `zio.schema.codec.DecodeError`, which is a sealed trait:

```scala
sealed trait DecodeErrorextends Exception with NoStackTrace { self =>
  def message
  override def getMessage(): String = message
  def and(that: DecodeError): DecodeError= DecodeError.And(self, that)
  def or(that: DecodeError): DecodeError= DecodeError.Or(self, that)
}
object DecodeError{
  final case class And(left: DecodeError, right: DecodeError) extends DecodeError
  final case class Or(left: DecodeError, right: DecodeError) extends DecodeError
  final case class MissingField(record: Record[_], field: Field[_], message: String) extends DecodeError
  final case class MalformedField(value: Schema[_], message: String) extends DecodeError
  final case class ReadError(cause: Cause[Any], message: String) extends DecodeError
  final case class ValidationError(validation: Validation[_], field: Field[_]) extends DecodeError
  final case class ExtraFields(fieldName: String, message: String) extends DecodeError
  ...
}
```

Then all decode methods can return `DecodingError` rather than `String`. Note that the exact subtypes of `DecodingError` should be determined by inspecting existing use cases across different protocols. The above is a guide and would likely be augmented with a few additional typed ways to fail.

devsprint:
Fixes #387


zio-schema/shared/src/main/scala/zio/schema/codec/DecodeError.scala
```diff
@@ -0,0 +1,45 @@
...
+    override def message: String = s"${left.message} or ${right.message}"
+  }
+
+  final case class RecordMissingField[R](record: Record[R], field: Field[R, _], message: String) extends DecodeError
+
+  final case class MissingField(value: Schema[_], message: String) extends DecodeError
+
+  final case class MalformedField(value: Schema[_], message: String) extends DecodeError
+
+  final case class MalformedFieldWithPath(path: Chunk[Any], message: String) extends DecodeError
```
jdegoes, line 36: Would be nice to make the `path` more well-typed here.

jdegoes: Or maybe go all the way to `String` if we can't do that.

jdegoes (approved):
Huge improvement!

Could we benefit from `EncodeError` now?

jdegoes:
@devsprint Good to merge when conflicts are fixed! We should think if adding `EncodeError` now makes sense.

zio-schema/shared/src/main/scala/zio/schema/codec/DecodeError.scala
devsprint: resolved, Changed to `Chunk[String]`

jdegoes (approved):

jdegoes:
Lint failure!

jdegoes (dismissed):

jdegoes:
Lint still failing

devsprint:
@jdegoes finally fixed.

PR #412, opened 2022-11-03
#164 Schema isomorphism for products

sviezypan:
New attempt at implementing schema isomorphism for product types. This PR contains refactoring of all CaseClassX schemas from `case class` to `sealed trait`.  The reason is increasing number of type parameters that we needed to support. Discussed in #373.

For example CaseClass2 is now a sealed trait with a bunch of fields like:
```
  sealed trait CaseClass2[A1, A2, Z] extends Record[Z] { self =>

    def field1: Field.WithFieldName[Z, Field1, A1]
    def field2: Field.WithFieldName[Z, Field2, A2]
    def construct: (A1, A2) => Z
    def annotations: Chunk[Any]
```

In addition every CaseClassX has `FieldX`, `Terms` and `FieldNames`
```
    type Field1 <: Singleton with String
    type Field2 <: Singleton with String    

    type Terms = (Field1, A1) with (Field2, A2)

    type FieldNames = Field1 with Field2
```

`FieldX` represents singleton type of field name, `AX` is a type of the field.

Every `CaseClassX` has also companion object that defines `apply` and `unapply` methods so existing code that manually created record schemas or pattern matched on schemas can remain unchanged. 

Macro derivation returns the following - so whoever wants, can take advantage of schema isomorphism
```
object CaseClass2 {
    type WithFields[F1 <: Singleton with String, F2 <: Singleton with String, A1, A2, Z] =
      CaseClass2[A1, A2, Z] {
        type Field1 = F1
        type Field2 = F2
      }
}
```

There is still some work left to do:
- scala 3 macro derivation returns just normal CaseClass, not the auxiliary type `WithFields`
- GenericRecord for classes of bigger arity than 22 also does not return schema with correct field names.

Still as this was big refactoring for record schemas, I was hoping we could already merge existing work in, and do the rest in the following PRs - I'll probably need little help with the above 2 points.


jdegoes:
@sviezypan I am ready to merge this quickly. There are a few conflicts due to other merges, can you please fix?

jdegoes:
@sviezypan There is one small change that I would like you to make before merging:

I see the lambdas are called `get0`, `field01`, etc., to avoid clashing with the methods. However, if you change the definition in the trait from a method to a lambda, then they can have the same name.

For example, change `Field` get from:

```
def get(s: S): A
```

to:

```
val get: S => A
```

This will allow us to use the same names, which improves the usability of the library because you don't have to remember the fields have a different name.

sviezypan:
hey @jdegoes, these conflicting fields are already defined as lambdas in traits.
```
def get0: R => A
def set0: (R, A) => R
```

To mimic your example, I tried to change `def` to `val` and renamed get0/set0 to get/set. It compiles but I am getting warnings
```
value get does nothing other than call itself recursively
value set does nothing other than call itself recursively
```

I believe this is not just a bogus message, because in tests the value of `get` and `set` lambdas in `Field` are null after macro derivation (tests are failing with NPE). So I assume we are calling `apply` with the correct values, they are just not propagated to the body when we create `Field`

```
    def apply[R, A](
      ....
      get: R => A,
      set: (R, A) => R
    ): Field[R, A] = new Field[R, A] {

      // other code
      val get: R => A               = get
      val set: (R, A) => R          = set
    }
```

jdegoes:
@sviezypan Try this:

```scala
    def apply[R, A](
      ....
      get0: R => A,
      set0: (R, A) => R
    ): Field[R, A] = new Field[R, A] {

      // other code
      val get: R => A               = get0
      val set: (R, A) => R          = set0
    }
```

sviezypan:
@jdegoes this is how its done right now. `get0` or 'field0' names are only parameters of `apply` method. Lambdas in traits are called `get` and `field`.

jdegoes:
@sviezypan Fantastic! One small conflict in `DeriveSchemaSpec`, and then good to merge!

PR #426, opened 2022-11-10
added caseNameAliases annotation with implementation in Json codec

### LINKED ISSUE #381: Implement support for @caseNameAliases annotation

This ticket is to implement the `@caseNameAliases` annotation. This annotation, which may be applied to any case of an enumeration (subtype of a `sealed trait`), contains a list of aliases the case is sometimes known by, which is a hint to the deserialization for different protocols to check for the case not only under its primary name, but also under its alternate names.

For example, maybe it may be the case that the `CreditCard` of `PaymentMethod` is usually stored in a field `CreditCard`, but occassionally appears under `credit_card`. Adding this annotation is a hint to the deserialization process to try not just `CreditCard`, but also, `credit_card`.

The main place where this new annotation should be used is in the different protocols that ZIO Schema supports. For example, the `zio-schema-json` protocol, which has serialization / deserialization in the JSON protocol.

ZIO Schema already captures annotations, so it is only necessary to look at the annotations and ensure that serialization / deserialization respects them in every protocol where the annotation makes sense.

devsprint:
Fixes #381


jdegoes (approved):

jdegoes:
@devsprint Good to merge when conflicts fixed!

jdegoes:
@devsprint Can you please fix conflicts? 🙏

jdegoes (approved):

Issue #419, opened 2022-11-07
Ability to capture implicit typeclass instances within the schema


vigoo:
# Problem

One of the core features zio-schema provides is to derive a `Schema[A]` for a type `A` once, and then use this data type to derive the implementation of additional functionalities such as codecs. Although the schema captures many information about the type - its structure, annotations, etc. - this might not be enough customization.

Consider for example the JSON codec in zio-schema. It is implemented in a way that it takes a `Schema[A]` and converts that to a zio-json codec:

```scala
def jsonCodec[A](schema: Schema[A]): ZJsonCodec[A]
```

normally with such a typeclass + derivation based design a user expects to be able to customize the instances, for example to implement a `ZJsonCodec` for a custom type by hand but use the typeclass derivation feature for the rest of the types. In this case the expectation is that the derivation macro picks up the hand-made implicit if it is in scope and does not even build the auto-generated codec for that particular type.

Using zio-schema to implement the derivation is a great simplification but it currently does not allow the above use case - once we captured the type's schema we no longer have enough type information to summon implicits for its inner types - the place where we could do that is the macro deriving the schema itself.

It would be nice to somehow be able to ask zio-schema to summon arbitrary typeclasses automatically so they can be readily accessed from the `Schema` types as a value. 

Detailed requirements:

- It should work separately from schema derivation - a 3rd party library may have predefined schemas (maybe even hand-written like in zio-flow) that we would like to reuse
- It should be something that takes a schema and a typeclass type and somehow annotates the schema's each node with summoned typeclass instances
- Transient fields should not summon the typeclass
- Should support annotating with more then once typeclasses, an (optional) optimization can be to do it in one run

# A draft solution proposal
One possible implementation could be a similar macro to the one that derives `Schema` itself and make it summon implicits and store them in the generated schema (similar to magnolia, but following the schema derivation macro's all features such as transient fields).

To minimize the required core changes it could store the summoned implicits in the schema's existing `annotations` fields. This would also allow storing multiple implicits, by providing a `T[_] => Annotation` mapping to this macro.

There are some open questions though:
- It feels right to say this macro takes an existing schema and annotates that. But in fact it has to replicate the type level traversal of the original macro in order to get the types for each level, so probably it does not really take much advantage of the provided existing schema. 
  - What if it is modified and there is a mismatch?
[...]

Issue #395, opened 2022-10-28
Reimplement the Scala 3 schema derivation without Mirror


vigoo:
Original problem:

The following does not work:

```scala
object T {
   implicit val schema: Schema[T] = DeriveSchema.gen
   val (...) = makeAccessors[T]
}

def makeAcessors[T](implicit schema: Schema[T]) = schema.makeAccessors(customAccessorBuilder)
```

while if we let the compiler to use the type generated by `DeriveSchema` it does:

```scala
object T {
   implicit val schema = DeriveSchema.gen[T]
   val (...) = makeAccessors[T]
}
```

The problem is that Scala 3 does not compile this, requiring all implicits to have explicit type annotation (which is the first version).

We could do the following trick to work around this:

```scala
object T {
   val generatedSchema = DeriveSchema.gen[T]
   implicit val schema: generatedSchema.type = generatedSchema
   val (...) = makeAccessors[T]
}
```

and this compiles on Scala 2, but it creates cyclic reference on Scala 3.

The reason for that is that `summon` is called _before_ checking whether we are on top level and that is easy to fix:

```scala
            val summoned = if (!top) Expr.summon[Schema[T]] else None
            summoned match {
              case Some(schema) =>
```
instead of
```scala
            Expr.summon[Schema[T]] match {
              case Some(schema) if !top =>
```

but that still does not work because `Mirror(typeRepr)` returns`None` for a type like that (I guess it is also trying to summon the implicit for `T` under the hood).

If the macro does not get the `Mirror` (just generates `null` for example) then the code with the above workaround compiles on Scala 3, so I believe a derive macro that does not rely on `Mirror` would solve this issue

PR #428, opened 2022-11-13
Typeclass derivation macro

vigoo:
Resolves https://github.com/zio/zio-schema/issues/419

This is a generic typeclass derivation macro that provides unified interface for Scala 2 and Scala 3 to derive any user defined type classes for data types having a `Schema` which is isomorphic to the schema generated by `DeriveSchema.gen`. 

Additional to the new type class derivation feature, it also contains several improvements to the Scala 3 version of the `DeriveSchema.gen` macro:
- It now properly refines the generated case class schemas with static field name information (`WithFields` etc)
- Support for case class and enum with arities higher than 22
- Support for `case object`
- Handles tuples the same way as Scala 2 (supporting only arities 2..4)
- I was able to delete the version specific (2.12, 2.13 and 3 separate) schema derivation test suites and have a single one
- Also seems to fix (with a version-specific implicit definition) the issue defined in https://github.com/zio/zio-schema/issues/395


zio-schema-derivation/shared/src/main/scala/zio/schema/CachedDeriver.scala
```diff
@@ -0,0 +1,188 @@
...
+  }
+
+  override def deriveTupleN[T](
+    schemasAndInstances: => Chunk[(Schema[_], WrappedF[F, _])],
+    summoned: => Option[F[T]]
+  ): F[T] =
+    throw new IllegalStateException(s"CachedDeriver uses explicit deriveTupleN overrides")
+}
+
+object CachedDeriver {
```
jdegoes, line 133: Can we make this whole file private and expose via `Deriver#cached`?

jdegoes (approved):

Issue #438, opened 2022-11-21
Allow for overriding the default Gen instance used for a type in zio-schema-zio-test


beem812:
Current behavior:
Gen instances for a string are automatically set to Gen.anyString

Problem: 
Gen.anyString can produce a character that postgres doesn't like, `\u0000`, which causes tests against a postgres db to randomly fail and do so extremely rarely making it harder to debug the cause.

Solution:
Some way to implicitly override the string gen for anything else, or a way to bake in the generator I want into a schema that can be used in place of the string primitive schema.

beem812:
I have a PR I'm working on that uses annotations to substitute the generator for a field if this seems like something we want. I've got it working but need to clean some things up, and clarify how amenable this project is to requiring a type tag in certain places.

PR #446, opened 2022-11-24
Override gen annotation

beem812:
`@generator()` annotation for specifying a different generator for the DeriveGen functionality to use. #438


build.sbt
```diff
@@ -109,7 +109,8 @@ lazy val zioSchema = crossProject(JSPlatform, JVMPlatform)
     libraryDependencies ++= Seq(
       "dev.zio" %% "zio"         % zioVersion,
       "dev.zio" %% "zio-streams" % zioVersion,
-      "dev.zio" %% "zio-prelude" % zioPreludeVersion
+      "dev.zio" %% "zio-prelude" % zioPreludeVersion,
+      "dev.zio" %% "zio-test"    % zioVersion
```
beem812, line 113: I had to include this for the annotation definition, would it make sense for this annotation to live inside the zio-schema-zio-test project to avoid including this?

vigoo:
If `DeriveGen` would be rewritten as a `Deriver` it would mean that:

- It can pick up custom implicit `Gen`s for a type
- That could be used to provide a custom `Gen[String]` for example

The annotation solution proposed in this PR is a bit more powerful because it allows customizing separate generators per usage within a single derivation, while with the implicit based you could just overwrite it once (but it could achieve the same thing by introducing multiple scoped implicits and expicitly deriving the `Gen` for subtrees).

zio-schema/shared/src/main/scala/zio/schema/annotation/generator.scala
```diff
@@ -0,0 +1,5 @@
+package zio.schema.annotation
```
jdegoes, line 1: Can we move this to the `zio-schema-zio-test` module?

jdegoes:
@vigoo All right, sounds like there may be a use case for this, then! Thanks.

@beem812 Just one comment, then good to merge after conflicts are fixed and build is passing. Thanks for your work on this!

vigoo:
@jdegoes anyway, rewriting `DeriveGen` with `Deriver` is a good idea (as a separate ticket) as it would enable the use of custom generators

build.sbt
jdegoes: Yes!

jdegoes:
@beem812 Happy to get this in. Please just re-open when conflicts are fixed and revisions are complete. Thank you!

PR #454, opened 2022-12-05
Extensible MetaSchema

### LINKED ISSUE #294: Ability to extend SchemaAst with known types
In zio-flow `ZFlow` and `Remote` are two types with a huge schema which is used everywhere recursively so we want to avoid serializing it in case of serializing schemas (using `SemiDynamic`). (Another reason to avoid serializing it is that it is currently broken as described in https://github.com/zio/zio-flow/issues/152)

To achieve this I had to copy a lot of code from zio-schema to https://github.com/vigoo/zio-flow/blob/serialization-fixes/zio-flow/shared/src/main/scala/zio/flow/serialization/FlowSchemaAst.scala in order to have special schema AST nodes representing these types in a compact way.

It would be nice to have support for this in zio-schema itself, possibly by having some schema registry and the opt-in possibility to serialize (or do any zio-schema operations) in a contextual way so the user can provide a schema registry.

vigoo:
Makes `MetaSchema` extensible with known types, which are represented by their `TypeId`. Support for `DynamicValue` in `MetaSchema` is reimplemented in top of this functionality.

Projects having some huge important built-in types and using `MetaSchema` to serialize custom types can use this feature to significantly reduce the size required by the serialized schemas. A good example is `zio-flow` where the `ZFlow` and `Remote` types (as well as `DynamicValue`) are very common in serialized schemas.

Resolves #294


zio-schema/shared/src/main/scala/zio/schema/meta/ExtensibleMetaSchema.scala
```diff
@@ -0,0 +1,724 @@
...
+  }
+
+  override def toString: String = AstRenderer.render(self)
+}
+
+object ExtensibleMetaSchema {
+  import CaseSet._
+
+  type Labelled[BuiltIn <: TypeList] = (String, ExtensibleMetaSchema[BuiltIn])
+  type Lineage                       = Chunk[(Int, NodePath)]
```
jdegoes, line 30: If these two are exposed to the outside world, may be better to make them `case class`.

vigoo: I agree. They were type aliases in the existing `MetaSchema` and I did not want to change more than necessary. Should I do it in this PR?

tests/shared/src/test/scala-2/zio/schema/MetaSchemaSpec.scala
```diff
@@ -123,28 +131,30 @@ object MetaSchemaSpec extends ZIOSpecDefault {
...
+                  "value1" -> ExtensibleMetaSchema
+                    .Value[DynamicValue :: End](StandardType.StringType, NodePath.root / "type2" / "value1"),
+                  "value2" -> ExtensibleMetaSchema.Product(
                     id = TypeId.parse("zio.schema.SchemaGen.Arity1"),
                     path = NodePath.root / "type2" / "value2",
                     fields = Chunk(
-                      "value" -> MetaSchema
-                        .Value(StandardType.IntType, NodePath.root / "type2" / "value2" / "value")
+                      "value" -> ExtensibleMetaSchema
+                        .Value[DynamicValue :: End](StandardType.IntType, NodePath.root / "type2" / "value2" / "value")
```
jdegoes, line 157: It may be helpful to have method-based constructors that can fill in `DynamicValue :: End` since it's so common. e.g. `def value(...): MetaSchema[...]`.

jdegoes: I mean, in the companion object of `ExtensibleMetaSchema`, of course.

vigoo: Almost did it :) but then I thought normal usage of `MetaSchema` does not really require dealing with its constructors, it is just this test suite. Normally you just use it as a serializable representation of a schema and not really manipulating it.

tests/shared/src/test/scala-2/zio/schema/MetaSchemaSpec.scala
```diff
@@ -253,6 +268,27 @@ object MetaSchemaSpec extends ZIOSpecDefault {
...
+        )
+      },
+      test("materializes the original schema") {
+        val meta1 = ExtendedMetaSchema.fromSchema(Schema[Pet])
+        val meta2 = ExtendedMetaSchema.fromSchema(Schema[DynamicValue])
+
+        val refEq1 = meta1.toSchema eq Schema[Pet]
+        val refEq2 = meta2.toSchema eq Schema[DynamicValue]
+        assertTrue(refEq1, refEq2)
+      }
```
jdegoes, line 291: There are no tests for the type list functionality. Could be useful to add them.

vigoo: These two test cases with `ExtendedMetaSchema` are trying to test that - they use `type ExtendedMetaSchema = ExtensibleMetaSchema[DynamicValue :: Pet :: End]` instead of the default. What else would you test?

zio-schema/shared/src/main/scala/zio/schema/meta/ExtensibleMetaSchema.scala
vigoo: Changed

jdegoes (approved):

PR #458, opened 2022-12-08
Override gen annotation for DeriveGen functionality

beem812:
`@generator()` annotation for specifying a different generator for the DeriveGen functionality to use. https://github.com/zio/zio-schema/issues/438

Allows for overriding the generator at both the field schema and the value schema level


jdegoes:
@beem812 I feel like the best way to support this would be to re-implement `DeriveGen` in terms of `Deriver`, which already allows implicits to override the default instances. What do you think?

beem812:
Sure, I'd be happy to take a look at it and attack it from that path. The lack of type safety from this solution felt weird anyway, hopefully the deriver route eliminates that issue.

PR #469, opened 2022-12-18
Fix discriminated json codec

paulpdaniels:
Fixes a bug in the json codec where the `caseName` and `discriminator` annotations couldn't be used together.

Added test cases to demonstrate the issue. I didn't look at protobuf as my current needs only require JSON but I'd imagine the same might need to be done there.

### DIFF
--- a/zio-schema-json/shared/src/main/scala/zio/schema/codec/JsonCodec.scala
+++ b/zio-schema-json/shared/src/main/scala/zio/schema/codec/JsonCodec.scala
@@ -276,42 +276,41 @@ object JsonCodec {
 
     private def enumEncoder[Z](parentSchema: Schema.Enum[Z], cases: Schema.Case[Z, _]*): ZJsonEncoder[Z] =
       (value: Z, indent: Option[Int], out: Write) => {
-        val nonTransientCases = cases.filter {
-          case Schema.Case(_, _, _, _, _, annotations) =>
-            annotations.collectFirst { case _: transientCase => () }.isEmpty
+        val nonTransientCase = cases.collectFirst {
+          case c @ Schema.Case(_, _, _, _, _, annotations)
+              if annotations.collectFirst { case _: transientCase => () }.isEmpty &&
+                c.deconstructOption(value).isDefined =>
+            c
         }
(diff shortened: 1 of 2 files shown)


zio-schema-json/shared/src/main/scala/zio/schema/codec/JsonCodec.scala
```diff
@@ -276,42 +276,41 @@ object JsonCodec {
...
     private def enumEncoder[Z](parentSchema: Schema.Enum[Z], cases: Schema.Case[Z, _]*): ZJsonEncoder[Z] =
       (value: Z, indent: Option[Int], out: Write) => {
-        val nonTransientCases = cases.filter {
-          case Schema.Case(_, _, _, _, _, annotations) =>
-            annotations.collectFirst { case _: transientCase => () }.isEmpty
+        val nonTransientCase = cases.collectFirst {
+          case c @ Schema.Case(_, _, _, _, _, annotations)
+              if annotations.collectFirst { case _: transientCase => () }.isEmpty &&
+                c.deconstructOption(value).isDefined =>
+            c
```
paulpdaniels, line 283: Previously this was iterating through the list at least twice (possibly a third as I'm not sure how `*` collections are backed when doing index lookups).

zio-schema-json/shared/src/main/scala/zio/schema/codec/JsonCodec.scala
```diff
@@ -462,12 +461,14 @@ object JsonCodec {
...
-                case a: caseNameAliases => a.aliases.toList
-                case cn: caseName       => List(cn.name)
-              }.getOrElse(List.empty))
+              annotations.flatMap {
+                case a: caseNameAliases => a.aliases.toList.map(_ -> name)
+                case cn: caseName       => List(cn.name -> name)
+                case _                  => Nil
+              }
+            case _ => Nil
           }.toMap
```
paulpdaniels, line 472: Inverting the `Map` to be `caseName -> name`, results in a possibly larger map but the `deAliasCaseName` will now be `O(1)`.

vigoo (approved):

vigoo:
Thanks!

Issue #447, opened 2022-11-29
Encoding Decoding Map to JSON


carlos-verdes:
How to reproduce, create a case class with an optional field of type Map:
```scala
case class ServerVersion(
    server: String,
    license: String,
    version: String,
    details: Map[String, String] = Map.empty
)

given Schema[ServerVersion] = DeriveSchema.gen[ServerVersion]
```

Expected test to pass:
```scala

      test("Encode/Decode ServerInfo from json using schema") {
        for
          _ <- ZIO.unit
          json = """{ "server": "arango", "license": "community", "version": "3.10.1"}"""
          chunks = Chunk.fromArray(json.getBytes)
          serverInfo <- ZIO.fromEither(
            JsonCodec.decode[ServerVersion](given_Schema_ServerVersion)(chunks)
          )
        yield assertTrue(
          serverInfo == ServerVersion("arango", "community", "3.10.1")
        )
      }
```

Error: the field details is decoded as `Map("" -> "" )`   instead of `Map.empty`:
![image](https://user-images.githubusercontent.com/8631522/204472252-6326d63e-1567-4d23-b0ba-8a5afd3b6211.png)

I tried also with next annotations:
```scala
import zio.schema.annotation.{fieldDefaultValue, optionalField}

case class ServerVersion(
    server: String,
    license: String,
    version: String,
    @optionalField
    @fieldDefaultValue[Map[String, String]](Map.empty)
    details: Map[String, String] = Map.empty
)
```

carlos-verdes:
I checked the Spec and I see Map is encoded as a sequence of tuples like:
```js
[
  {"key1":  value1},
  {"key2":  value2}
]
```

However the expectation for me is:
```js
{
  "key1":  value1,
  "key2":  value2
}
```

I think we should change the full implementation of Map encoder/decoder not just the default values

carlos-verdes:
@jdegoes @thinkharderdev @devsprint  @tusharmath 

This is a big problem to use in any API, I have to receive/pass very often Map[String, String] for extra configuration (for example is you use Spark) and it's really weird that is mapped to json [] instead of {}

I implemented a library for ArangoDB using zio-http and zio-json for serialization and I'm now trying to support zio-schema on top of that (so you can use ArangoDB with any type that has an Schema, and I plan to do dynamic graph queries that bring more than 1 entity just from the Schema abstraction)

Andrapyre:
@adamgfraser , maybe you want to claim a bounty for this?

[...]

PR #502, opened 2023-01-30
Improvements to the schema based Json codec

vigoo:
- Annotation for not using any discriminators for enums. This can be used to encode things like OpenAPI `oneOf` types. The decoder in this mode tries decoding each case until one succeeds.
- Annotation for directly mapping `DynamicValue` to JSON. This mode does not support everything DynamicValue does, but allows us to have dynamic sub-JSONs in our generated codecs. It can also be a workaround for the issue https://github.com/zio/zio-schema/issues/447
- Fixes an issue that if an optional field was wrapped in a `Transform`, it was no longer handled correctly in the JSON codec (omitting vs writing explicit null)

### DIFF
--- a/zio-schema/shared/src/main/scala/zio/schema/annotation/directDynamicMapping.scala
+++ b/zio-schema/shared/src/main/scala/zio/schema/annotation/directDynamicMapping.scala
@@ -1,0 +1,11 @@
+package zio.schema.annotation
+
+import scala.annotation.StaticAnnotation
+
+/**
+ * When applied to a Schema.Dynamic schema, indicates that the dynamic value should be directly mapped to the target codec
+ * if possible.
+ *
+ * For example the JSON codec can encode DynamicValue.Record to a JSON object directly instead of encoding the DynamicValue structure.
+ */
+final case class directDynamicMapping() extends StaticAnnotation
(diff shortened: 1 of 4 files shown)


zio-schema/shared/src/main/scala/zio/schema/annotation/directDynamicMapping.scala
```diff
@@ -0,0 +1,11 @@
...
+
+import scala.annotation.StaticAnnotation
+
+/**
+ * When applied to a Schema.Dynamic schema, indicates that the dynamic value should be directly mapped to the target codec
+ * if possible.
+ *
+ * For example the JSON codec can encode DynamicValue.Record to a JSON object directly instead of encoding the DynamicValue structure.
+ */
+final case class directDynamicMapping() extends StaticAnnotation
```
jdegoes, line 11: Could be `nativeDynamicMapping` maybe?

jdegoes (approved):

carlos-verdes:
The enum without discriminator also is a feature I was looking for!
Nice PR

carlos-verdes:
@vigoo how can I use this as a workaround for #447 ?

PR #517, opened 2023-02-23
Initial support for opaque types in Deriver

### LINKED ISSUE #507: Deriver fails at runtime on opaque types
When using an opaque that is defined in another object (in the same object, it works because the opaque type is equivalent to its inner type), the `Deriver` fails with:
```
Cannot derive instance for type java.lang.String
	at zio.schema.Deriver.deriveUnknown(Deriver.scala:100)
	at zio.schema.Deriver.deriveUnknown$(Deriver.scala:33)
	at zio.schema.Test$$anon$1.deriveUnknown(Test.scala:19)
```

Reproducer:
```scala
package zio.schema

import zio.Chunk

object AnotherObject {
  opaque type MyId = String

  object MyId {
    def apply(s: String): MyId                     = s
    given (using ev: Schema[String]): Schema[MyId] = ev
  }
}

object Test extends App {
  trait TC[A] {
    def name(a: A): String
  }

  val deriver = new Deriver[TC] {
    override def deriveRecord[A](record: Schema.Record[A], fields: => Chunk[Deriver.WrappedF[TC, _]], summoned: => Option[TC[A]]): TC[A] = ???

    override def deriveEnum[A](`enum`: Schema.Enum[A], cases: => Chunk[Deriver.WrappedF[TC, _]], summoned: => Option[TC[A]]): TC[A] = ???

    override def derivePrimitive[A](st: StandardType[A], summoned: => Option[TC[A]]): TC[A] = ???

    override def deriveOption[A](option: Schema.Optional[A], inner: => TC[A], summoned: => Option[TC[Option[A]]]): TC[Option[A]] = ???

    override def deriveSequence[C[_], A](sequence: Schema.Sequence[C[A], A, _], inner: => TC[A], summoned: => Option[TC[C[A]]]): TC[C[A]] = ???

    override def deriveMap[K, V](map: Schema.Map[K, V], key: => TC[K], value: => TC[V], summoned: => Option[TC[Map[K, V]]]): TC[Map[K, V]] = ???

    override def deriveTransformedRecord[A, B](
                                                record: Schema.Record[A],
                                                transform: Schema.Transform[A, B, _],
                                                fields: => Chunk[Deriver.WrappedF[TC, _]],
                                                summoned: => Option[TC[B]]
                                              ): TC[B] = ???
  }

  given TC[AnotherObject.MyId] = deriver.derive

  def show[A](a: A)(using ev: TC[A]): Unit =
    println(ev.name(a))

  show(AnotherObject.MyId("test"))
}
```

In the deriver, the opaque type is not recognized as a `StandardType` nor a case class or a sealed trait, so it ends up in `deriveUnknown` and fails there. The `Schema` is not even used.

Note that it also fails if the opaque type is not the top object.
I ran into this issue trying to use Desert with opaque types.

vigoo:
Resolves #507


zio-schema-derivation/shared/src/test/scala-3/zio/schema/VersionSpecificDeriveSpec.scala
```diff
@@ -0,0 +1,70 @@
...
+      def apply(s: String): MyId = s
+
+      given(using ev: Schema[String]): Schema[MyId] = ev
+    }
+  }
+
+  trait TC[A] {
+    def name(a: A): String
+  }
+}
```
kajebiii, line 70: Can you add newline at end of this file ?

PR #531, opened 2023-03-22
Update Dependencies

adamgfraser:


vigoo (approved):

carlos-verdes:
When is this upgrade planned to be deploy?
I'm kind of block now after I upgraded zio dependencies

Explanation of my problem:
I have zio-schema version 0.4.9 which still uses zio-json 0.3.0 and I have in context zio-json 0.5.0 (which is not compatible with zio-schema 0.4.9)

vigoo:
Started a release now

carlos-verdes:
Cool, can you take a look at my latest PR #534?
It's related with issue #447 

It's something I need for long time, current implementation Encodes Map[String, Any] like a JSON array, my expectation is to be rendered as a JSON Object.

Example:
```scala
  val config: Map[String, String] = Map("extra" -> "config")
  val expectedJson = """{"extra":"config"}"""

  val currentImplementation = "[{"extra":"config"}]
```

Issue #519, opened 2023-03-02
Derivation Error in Scala 3.2.2


tusharmath:
Observed the following error —

> `Unreachable case except for null (if this is intentional, consider writing case null => instead).`

```scala
import zio.schema._

sealed trait Expression

object Expression {
  implicit val schema: Schema[Expression] = DeriveSchema.gen[Expression]
  case object Identity extends Expression
}
```

This happens when `"-Xfatal-warnings"` is set in scalacOptions. 

Here is a small repro of the issue — 

https://scastie.scala-lang.org/5Sau1ZC5R0yEwpfS1kybOA

PR #545, opened 2023-04-25
Fix i519

KacperFKorban:
closes zio#519

/claim #519

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -124,7 +124,6 @@ lazy val zioSchemaDerivation = crossProject(JSPlatform, JVMPlatform)
   .in(file("zio-schema-derivation"))
   .dependsOn(zioSchema, zioSchema % "test->test")
   .settings(stdSettings("zio-schema-derivation"))
-  .settings(dottySettings)
   .settings(crossProjectSettings)
   .settings(buildInfoSettings("zio.schema"))
   .settings(
@@ -142,15 +141,6 @@ lazy val zioSchemaDerivation = crossProject(JSPlatform, JVMPlatform)
             "org.scala-lang" % "scala-reflect" % scalaVersion.value % Provided
           )
         case _ => Seq()
-      }
-    }
-  )
-  .settings(
-    Test / scalacOptions ++= {
-      CrossVersion.partialVersion(scalaVersion.value) match {
-        case Some((3, _)) =>
-          Seq("-Wconf:name=MatchCaseOnlyNullWarning:e")
-        case _ => Seq.empty
       }
     }
   )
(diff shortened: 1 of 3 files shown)


build.sbt
```diff
@@ -145,6 +145,15 @@ lazy val zioSchemaDerivation = crossProject(JSPlatform, JVMPlatform)
...
   )
+  .settings(
+    Test / scalacOptions ++= {
+      CrossVersion.partialVersion(scalaVersion.value) match {
+        case Some((3, _)) =>
+          Seq("-Wconf:name=MatchCaseOnlyNullWarning:e")
+        case _ => Seq.empty
+      }
+    }
+  )
```
tusharmath, line 156: Might want to remove the `scalacOptions -= "-Xfatal-warnings"` instruction that's added at a few places in the `build.sbt` file

jdegoes:
@tusharmath Is this good to merge?

build.sbt
KacperFKorban: Ok, removed this custom setting and removed `dottySettings` from `zioSchemaDerivation` instead.

tusharmath:
No, it still fails with the following error —

```
[error] -- [E121] Pattern Match Error: /Users/tushar/Documents/Projects/zio-schema/zio-schema-examples/shared/src/main/scala/dev/zio/schema/example/Foo.scala:8:53 
[error]   8 |  implicit val schema: Schema[Foo] = DeriveSchema.gen[Foo]
[error]     |                                     ^^^^^^^^^^^^^^^^^^^^^
[error]     |Unreachable case except for null (if this is intentional, consider writing case null => instead).
[error]     |---------------------------------------------------------------------------
[error]     |Inline stack trace
[error]     |- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
[error]     |This location contains code that was inlined from DeriveSchema.scala:461
[error] 461 |          case other => throw new MatchError(other)
[error]     |               ^^^^^
[error]     |- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
[error]     |This location contains code that was inlined from DeriveSchema.scala:461
[error] 458 |      val unsafeDeconstruct = '{ 
[error]     |                              ^
[error] 459 |        (z: T) => z match {
[error] 460 |          case (sub: t) => sub
[error] 461 |          case other => throw new MatchError(other)
[error] 462 |        }
[error] 463 |       }
[error]      ---------------------------------------------------------------------------
[error] one error found

```


**[repro.patch](https://github.com/zio/zio-schema/files/11339511/repro.patch)**

- Apply the attached patch to your revision.
- Run the following command `reload; clean; ++3.2.1;zioSchemaExamplesJVM/compile`
- **Expected** — code should compile
- **Actual** - code doesn't compile.

tusharmath:
Good to merge!

tusharmath (approved):

PR #549, opened 2023-04-30
Rewrite DynamicValue error handling to Validate

### LINKED ISSUE #480: Improve DynamicValue error handling
In current implementation, `toTypedValue` operation is returning an Either data type, that is basically an implementation of `fail-fast` approach on error handling.
Instead, if could return an instance of `Validation` data type from `zio-prelude`, to accumulate all the errors in the process of transforming a `DynamicValue` instance into a typed value.

andrzejressel:
Due to the fact that most if the zio-schema codebase is using Either, there must be `.toEither.left.map(_.toString)` on the edge.

I've tried to going further than that, but it would require changes in the whole codebase.

Requires: https://github.com/zio/zio-json/issues/771

Fixes #480
/claim #480


jdegoes:
@andrzejressel I modified the bounty to $150, can you push `Validation` all the way through so we can benefit from these accumulated errors past the edge?

andrzejressel:
Fortunately it does not look that bad in diff view. I have to rewrite examples and add do changes on zio-json and zio-bson

andrzejressel:
Since I don't have so much time anymore and ZIO-JSON seems to not like accumulation: https://discord.com/channels/629491597070827530/733728086637412422/1113027920420675674 - I'm closing this MR.

Issue #478, opened 2023-01-15
Add scaladoc to codec annotations


googley42:
Issue https://github.com/zio/zio-schema/issues/203 has been implemented but it looks like there are missing scaladoc for the annotations:

> ... add and document these annotations inside zio.schema.annotation ...

iamorozov:
I would like to try this one

987Nabil:
@jdegoes @vigoo actually, for Scala 3 we can extract Scala doc at macro execution. But this ticket is now already rewarded and has a merged pr. Should we open a new pr?

vigoo:
If we can capture it it would definitely be a nice feature

Issue #203, opened 2022-02-18
Define a canonical set of annotations that can be applied to sealed traits, case classes, and fields


jdegoes:
Downstream libraries sometimes want to offer the ability to tweak the way codecs are generated from case classes, sealed traits, and fields inside case classes. A convenient way to do this is with annotations, e.g. ZIO DynamoDB's `@discriminator("type")` annotation, which specifies a different way to encode sealed trait hierarchies.

Rather than encouraging library authors to come up with a set of per-library annotations, with attendant duplication, we should instead introduce a "canonical" set of annotations that we expect can be used by a wide percentage of downstream library authors.

I'd start off by proposing the following set of annotations, which could live inside `zio.schema.annotation._`, and which could offer a standardized way to tweak common codec settings.

 - [ ] `@discriminatorName`, for specifying the name of a field that will be used to contain the id which identifies which term in an enum is being serialized / deserialized. For example, if you set `@discriminatorName("type")`, then a field called "type" will be used to store the identity of the case of the enum that is being serialized / deserialized.
 - [ ] `@fieldName`, for specifying an alternate identity for a field of a case class. Currently, the field name is always used. But sometimes, it will be convenient to have control over that name. For example, maybe the API expects username to be stored as `user_name`, but inside the case class, the field is named `username`. Such an annotation, applied directly to the field, could indicate a different identity for the field than the field name itself.
 - [ ] `@caseName`, for specifying an alternate identity for a case of an enum. Currently, the subtype name is always used. This is the dual of `@fieldName`, but for sums rather than products.
 - [ ] '@typeName`, for specifying an alternate name for a data type (either a case class or sealed trait).
 - [ ] `@fieldNameAliases("alias1", "alias2")`, for specifying a list of aliases the field is sometimes known by.
 - [ ] `@caseNameAliases("alias1", "alias2")`, for specifying a list of aliases the case is sometimes known by.
 - [ ] `@optionalField`, specifying that deserialization should not fail even if a value of this type is not specified.
 - [ ] `@rejectExtraFields`, specifying that deserialization should reject payloads that contain more fields than specified.
 - [ ] `@transientField`, specifying that serialization should make no effort to encode the field to which the annotation is applied.

This ticket is only to add and document these annotations, not to actually use them in the codecs provided by _ZIO Schema_. That work can be specified and completed in other, separate tickets.

vigoo:
`@transientCase` would also be useful to ignore some cases of a sum type completely (that are known to be never serialized)

[...]

PR #550, opened 2023-05-01
Add scaladoc to annotations

iamorozov:
This PR adds scaladoc formatted documentation to annotations in `zio.schema.annotation` package implemented in https://github.com/zio/zio-schema/issues/203

/claim #478

### DIFF
--- a/zio-schema/shared/src/main/scala/zio/schema/annotation/caseName.scala
+++ b/zio-schema/shared/src/main/scala/zio/schema/annotation/caseName.scala
@@ -2,4 +2,12 @@ package zio.schema.annotation
 
 import scala.annotation.StaticAnnotation
 
+/**
+ * Annotation for specifying an alternate identity for a case of an enum.
+ * Currently, the subtype name is always used.
+ *
+ * This is the dual of `@fieldName`, but for sums rather than products.
+ *
+ * @param name The alternate name to use for the case.
+ */
 final case class caseName(name: String) extends StaticAnnotation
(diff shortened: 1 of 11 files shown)


jdegoes:
This is a great addition, however, it's not exactly what the ticket is looking for.

The ticket is looking for modifying the macros that auto-generate `Schema` (the derivation macros), and incorporating the Scaladoc attached to the user's own `case class` and `sealed trait` data types, into the `Schema` class itself, possibly as an annotation.

So, for example, if you had:

```scala
/**
 * My Scaladoc.
 */
final case class Person(name: String, age: Int)
```

Then you auto-derived `Schema[Person]` by using `DeriveSchema.gen[Person]`, then if you looked into the annotations on `Schema[Person]`, you would find one that contains the Scaladoc, including the string `My Scaladoc.`.

Does this make sense? Are you still interested in implementing this ticket?

iamorozov:
@jdegoes Thank you for clarifying the task, it is more clear now and makes sense. I will try to make all changes needed to achieve this

iamorozov:
@jdegoes As scaladoc is not being included into compiled classes and thus not extractable with reflection I wonder what approach should I use to add scaladoc to derived schema. Is extracting and parsing source file sounds applicable for this task?

vigoo:
@jdegoes I think the original linked issue was really just talking about adding documentation comments in the code for the annotations (as the PR does). I would not try to parse the source code separately from the macro, if the doc comments are not provided by the compiler.

jdegoes:
@vigoo All right, happy to go with your interpretation (it's less work, anyway).

I had assumed that Scaladoc would be visible at compile-time to macros and not require source parsing.

vigoo (approved):

PR #555, opened 2023-05-08
Scala 3: don't make the self ref vals private within

KacperFKorban:
When creating symbols, their `privateWithin` has to be a type.
When defining `zio-schema` derivation e.g.

```scala
given x: Schema[ArangoError] = DeriveSchema.gen[ArangoError]
```

`Symbol.spliceOwner` of the derivation macro was `x`. Which is not a correct value for `privateWithin`.

This property will now be enforced with https://github.com/lampepfl/dotty/pull/17352

EDIT: Also after this fix `-Xprint` should work for programs using `zio-schema` derivations.


jdegoes:
Thank you! Can you run `sbt docs/generateReadme` to update the README, then the CI build can make further progress!

KacperFKorban:
@jdegoes rebased, the CI is green now

jdegoes (approved):

PR #556, opened 2023-05-10
Support discriminator in any place in object

### LINKED ISSUE #523: JSON decoding failure for ADTs
Given an ADT and a decoding program given below

```scala
@discriminatorName("type")
sealed trait IOInteraction

object IOInteraction {
  @caseName("http")
  case class Sync(url: String) extends IOInteraction

  @caseName("kafka")
  case class Async(topic: String) extends IOInteraction

  given Schema[IOInteraction] = DeriveSchema.gen[IOInteraction]
}


@main def MainApp = {
  def decode[A: Schema](input: String) = JsonDecoder.decode(summon[Schema[A]], input)
  val result1 = decode[IOInteraction](syncStr)
  println(result1)
}
```

it works successfully with this input string:

```scala
val syncStr = """{ "type": "http" ,  "url": "http://localhost" }"""
```

but not with

```scala
val syncStr = """{ "url": "http://localhost" , "type": "http"  }"""
```

The only diff is order of discriminator field.

I am getting below error

```
Left(zio.schema.codec.DecodeError$ReadError: .url(unrecognized subtype))
```

From [discord chat](https://discord.com/channels/629491597070827530/822172415026790531/1084868104338804856)

andrzejressel:
I wish there was better ways than `replay`, but since we have to move through entire object, I don't think there is other way.

Fixes #523 
/claim #523

### DIFF
--- a/zio-schema-json/shared/src/main/scala/zio/schema/codec/JsonCodec.scala
+++ b/zio-schema-json/shared/src/main/scala/zio/schema/codec/JsonCodec.scala
@@ -484,56 +484,56 @@ object JsonCodec {
     }
 
     //scalafmt: { maxColumn = 400, optIn.configStyleArguments = false }
-    private[codec] def schemaDecoder[A](schema: Schema[A], hasDiscriminator: Boolean = false): ZJsonDecoder[A] = schema match {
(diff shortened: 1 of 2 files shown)


jdegoes:
@andrzejressel Unfortunately, `replay` will slow things down tremendously. It would be better to lazily discover the discriminator when moving through the tokens of the JSON object.

Issue #547, opened 2023-04-29
Compilation error on recursive GADT


senia-psm:
This code fails compilation:

```scala
sealed trait GenericTree[T]

object GenericTree {
  case class Branch[T](left: GenericTree[T], right: GenericTree[T]) extends GenericTree[T]
  case class Leaf[T](value: T) extends GenericTree[T]

  implicit def schema[T: Schema]: Schema[GenericTree[T]] = DeriveSchema.gen[GenericTree[T]]
//    implicit lazy val schemaInt: Schema[GenericTree[Int]] = DeriveSchema.gen[GenericTree[Int]]
}
```

Replacing generic `schema` with concrete `schemaInt` does not solve the issue.

```
[error] scala.ScalaReflectionException: object scala.reflect.macros.blackbox.Context in compiler mirror not found.
[error]         at scala.reflect.internal.Mirrors$RootsBase.staticClass(Mirrors.scala:145)
[error]         at scala.reflect.internal.Mirrors$RootsBase.staticClass(Mirrors.scala:29)
...
[error]         at zio.schema.DeriveSchema$.genImpl(DeriveSchema.scala:644)
[error]     implicit def schema[T: Schema]: Schema[GenericTree[T]] = DeriveSchema.gen[GenericTree[T]]
```

This code works just fine:

```scala
sealed trait GenericTree[T]

object GenericTree {
  case class Branch[T](left: GenericTree[T], right: GenericTree[T]) extends GenericTree[T]

  object Branch {
    implicit def schema[T: Schema]: Schema[Branch[T]] = DeriveSchema.gen[Branch[T]]
  }

  case class Leaf[T](value: T) extends GenericTree[T]

  object Leaf {
    implicit def schema[T: Schema]: Schema[Leaf[T]] = DeriveSchema.gen
  }

  implicit def schema[T: Schema]: Schema[GenericTree[T]] = DeriveSchema.gen[GenericTree[T]]
}

```

PR #561, opened 2023-05-19
Fix compilation of Recursive GADT schema derivation

Jesse-Bakker:
/claim #547


Jesse-Bakker:
@zio/zio-schema @jdegoes gentle reminder that this needs review, thanks!

PR #562, opened 2023-05-20
Read scala default values, if no fieldDefaultValue annotation is found

### LINKED ISSUE #552: Get default value from case class default paramters
It's very convenient to define default value in case class, and zio-json do so.

987Nabil:
fixes #552
/claim #552

### DIFF
--- a/zio-schema-derivation/shared/src/main/scala-3/zio/schema/DeriveSchema.scala
+++ b/zio-schema-derivation/shared/src/main/scala-3/zio/schema/DeriveSchema.scala
@@ -275,13 +275,40 @@ private case class DeriveSchema()(using val ctx: Quotes) extends ReflectionUtils
         field.name -> field.annotations.filter(filterAnnotation).map(_.asExpr)
     }
 
-  private def fromConstructor(from: Symbol): scala.collection.Map[String, List[Expr[Any]]] =
+  private def defaultValues(from: Symbol): Predef.Map[String, Expr[Any]] =
+    (1 to from.primaryConstructor.paramSymss.size).toList.map(
+      i =>
+          from
+            .companionClass
+            .declaredMethod(s"$$lessinit$$greater$$default$$$i")
+            .headOption
+            .orElse(
+              from
+                .companionClass
+                .declaredMethod(s"$$apply$$default$$$i")
+                .headOption
+            )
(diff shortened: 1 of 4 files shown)


liewhite:
Any progress?

jdegoes (dismissed):

987Nabil:
@jdegoes I fixed the linting errors

liewhite:
I test this locally, it failed on generic field.

987Nabil:
@liewhite Can you provide a reproducer and tell me for which Scala version?

zio-schema-derivation/shared/src/main/scala-3/zio/schema/DeriveSchema.scala
```diff
@@ -275,13 +275,40 @@ private case class DeriveSchema()(using val ctx: Quotes) extends ReflectionUtils
...
+            .companionClass
+            .declaredMethod(s"$$lessinit$$greater$$default$$$i")
+            .headOption
+            .orElse(
+              from
+                .companionClass
+                .declaredMethod(s"$$apply$$default$$$i")
+                .headOption
+            )
+            .map(Select(Ref(from.companionModule), _).asExpr)
```
liewhite, line 291: Compiling failed on Generic case class here.

liewhite:
> @liewhite Can you provide a reproducer and tell me for which Scala version?

```scala
case class A[T](a: Option[T], b: Int) derives Schema
```
Compilation failed with `not a Expr` error. 

I comment in code view

987Nabil:
Thanks. I'll take a look soon.

liewhite:
> Thanks. I'll take a look soon.
```scala
if(i.isExpr) {
        i.asExpr
      }else {
        i.appliedToType(TypeRepr.of[Any]).asExpr
      }
```
This works.
But is it safe to use `any` after compiler type check?

987Nabil:
@liewhite Finally found some time. Your solution is ok, since when we call `toExpr` it ends up being an `Expr[Any]` anyway.
But there have been more problems I needed to fix. Now it looks ok.

@jdegoes I could not use a generic implicit for the test of the generic class. If I do it, the gen macro for Scala 3 fails. But this seems to be a problem that existed before my changes. I'll open a bg ticket.

vigoo (dismissed):

vigoo (approved):

Issue #506, opened 2023-02-02
add a codec annotation to mark a sealed trait as having all case object members


googley42:
in zio-dynamodb (which is uses zio-schema) we have the `@enumOfCaseObjects` trait level annotation to signify that all members as case objects. This gets read by our codec generation code to produce a more compact encoding that is used in many legacy databases. 

eg given this model:

```scala
sealed trait TrafficLight
case object GREEN extends TrafficLight 
case object RED extends TrafficLight
final case class Box(trafficLightColour: TrafficLight)
```
The default tagged union encoding is something like:
```
Map(trafficLightColour -> Map(String(GREEN) -> Null))
``` 
after you add the annotation to the sealed trait, the encoding is more compact, with one less level of indirection:
```scala
Map(trafficLightColour -> String(GREEN))
```

This issue is related to this one [Define a canonical set of annotations that can be applied to sealed traits, case classes, and fields](https://github.com/zio/zio-schema/issues/203)

jdegoes:
Maybe `@SimpleEnum` to indicate all terms are `case object`s.

andrzejressel:
Weird, then looking at schema recently I swore I've seen code like that - if everything is sealed trait is CaseObject0 it deserialises as single string. I'll take a look later.

PR #564, opened 2023-05-24
added simpleEnum annotation

pablf:
/claim #506
Throws exception if used where not applicable and it is added automatically if it is indeed a simple Enum.

### DIFF
--- a/zio-schema-derivation/shared/src/main/scala-2/zio/schema/DeriveSchema.scala
+++ b/zio-schema-derivation/shared/src/main/scala-2/zio/schema/DeriveSchema.scala
@@ -528,7 +528,7 @@ object DeriveSchema {
             case annotation =>
               c.warning(c.enclosingPosition, s"Unhandled annotation ${annotation.tree}")
               EmptyTree
-          }.filter(_ != EmptyTree).prepended(q"new _root_.zio.schema.annotation.simpleEnum()")
+          }.filter(_ != EmptyTree).+:(q"new _root_.zio.schema.annotation.simpleEnum(true)")
         case (false, true) =>
           c.abort(c.enclosingPosition, s"${show(tpe)} must be a simple Enum")
         case _ =>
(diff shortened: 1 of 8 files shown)


googley42:
@pablf so is the intention to automatically insert the `simpleEnum` into the derived schema? I think in the original issue I wanted to control the application of the annotation (ie make this manual process) so that users of downstream libraries like `zio-dynamodb` can choose the encoding  they want eg new projects may want the default tagged union encoding and projects that need to integrate with legacy databases say could use this new annotation.

pablf:
And if we included an `automaticallyAdded: Boolean` in the annotation's definition? What do you think?

pablf:
Now it should pass the tests.
@googley42 I have added automaticallyAdded. If you want to use you can use `@simpleEnum` and it will generate a `simpleEnum(false)` annotation.

googley42:
@pablf  so I think the original intention of the issue was just to create a standard annotation for use by downstream libraries - we already have one for zio-dynamodb - but the the thought was to standardise it. The fact that the JSon codec already defaults to this behaviour probably means that there is no extra work required for derive gen macro - what do you think @jdegoes ?

googley42:
@pablf 

> Maybe it would also be interesting for a Schema to know if it is a simple Enum, so each one of its codecs can use that information as desired, instead of checking in each codec if it is really a simple Enum

Ah yes - I get this now - this is _great_ idea and will simplify the codecs of all downstream libs!

The only thing I want to confirm is that will a manually created `@simpleEnum` override the automatically generated one? I suspect this is case and if so do you think it would be a good idea to have a unit test for this?

pablf:
@googley42 Then great! Yes, that would be the expected behaviour. I have added the tests to ensure that it's like that in the future.

googley42 (approved):
LGTM and I think this a great piece of functionality - however I am not a zio-schema maintainer or have much knowledge around macros! you may want to get another review cc @zio/zio-schema

googley42 (approved):
.

vigoo (approved):

vigoo (approved):

PR #587, opened 2023-08-02
upgrade Scala 3 to 3.3

alphaho:


zio-schema-derivation/shared/src/main/scala-3/zio/schema/Derive.scala
```diff
@@ -14,7 +14,9 @@ object Derive {
     DeriveInstance().deriveInstance[F, A](deriver, schema, top = true)
 }
 
-private case class DeriveInstance()(using val ctx: Quotes) extends ReflectionUtils(ctx) {
```
alphaho, line 17: Need to move away from extending `ReflectionUtils`, or it will get `The type of a class parent cannot refer to constructor parameters, but zio.schema.ReflectionUtils[(DeriveInstance.this.ctx : quoted.Quotes)] refers to ctx`

vigoo:
Updated as part of another PR

PR #605, opened 2023-09-06
Ways to ignore empty collections while encoding json

987Nabil:
This pr introduces a config for the JsonCodec, to set behaviour on the codec creation level instead of classes via annotations. We can use this to configure default behaviour in zio-http for example via user config.
Or could make it the default of zio-http to ignore empty collection fields during encoding, without messing with the defaults of zio-schema.

### DIFF
--- a/zio-schema-json/shared/src/main/scala/zio/schema/codec/JsonCodec.scala
+++ b/zio-schema-json/shared/src/main/scala/zio/schema/codec/JsonCodec.scala
@@ -23,6 +23,13 @@ import zio.stream.ZPipeline
 import zio.{ Cause, Chunk, ChunkBuilder, NonEmptyChunk, ZIO }
 
 object JsonCodec {
+
+  final case class Config(ignoreEmptyCollections: Boolean)
+
+  object Config {
+    val default: Config = Config(ignoreEmptyCollections = false)
+  }
+
   type DiscriminatorTuple = Chunk[(discriminatorName, String)]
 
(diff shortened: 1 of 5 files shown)


vigoo (dismissed):

jdegoes:
@987Nabil Overall looks good. I am not a fan of the implicit config pattern. Do we have to make that config implicit?

987Nabil:
@jdegoes I guess we can have overloaded methods with a config parameter. So we don't break compatibility

zio-schema-json/shared/src/main/scala/zio/schema/codec/JsonCodec.scala
```diff
@@ -58,7 +65,7 @@ object JsonCodec {
         )
     }
 
-  implicit def schemaBasedBinaryCodec[A](implicit schema: Schema[A]): BinaryCodec[A] =
+  implicit def schemaBasedBinaryCodec[A](implicit schema: Schema[A], cfg: Config): BinaryCodec[A] =
```
vigoo, line 68: Now `Config.default` is no longer implicit, but here it is an implicit parameter which makes this a bit inconvenient. I suggest to keep an `implicit def schemaBasedBinaryCodec` with only schema parameter, always using `Config.default`, so users who want to customize it can use the explicit version.

987Nabil: Ahh sorry, I missed that part

vigoo (dismissed):

vigoo (approved):

PR #611, opened 2023-10-20
Find implicits first, before deriving any kind of schema in Scala 3

987Nabil:
Currently, the macro for Scala 3 is ignoring self defined schemas for collections.
So if for example a `implicit val s: Schema[Map[X,Y]]` is in scope and you derive a schema for a case class that has a field of that type, the macro derives a map schema for derived X & Y schemas instead of using the implicit.
That is imho a bug, since user defined schemas should always be preferred.


987Nabil:
this is btw already the case for the Scala 2 macro. So this aligns the behaviour of Scala 3 with Scala 2 derivation.

adamgfraser (approved):

PR #623, opened 2023-12-22
Fix typo in StandardType.Tags.LOCAL_DATE_TIME

### LINKED ISSUE #620: Typo in string literal
Hi there.
Please fix "I" to "i" in "localDateTIme" :)

https://github.com/zio/zio-schema/blob/82eace0bcbe5a89fde8162ba88bd3d54dbc4e9fe/zio-schema/shared/src/main/scala/zio/schema/StandardType.scala#L49

Best Regards

alphaho:
Fix #620

### DIFF
--- a/zio-schema/shared/src/main/scala/zio/schema/StandardType.scala
+++ b/zio-schema/shared/src/main/scala/zio/schema/StandardType.scala
@@ -46,7 +46,7 @@ object StandardType {
     final val INSTANT          = "instant"
     final val LOCAL_DATE       = "localDate"
     final val LOCAL_TIME       = "localTime"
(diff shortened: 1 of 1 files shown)


987Nabil:
Is this actually a breaking change? If I have a meta schema in a db,  has the typo and we fix this,  deserialization will fail. I understand that it is not beautiful, but imho the guarantee that  we need to keep is, that for everything type we have a fixed type tag. 
I would not fix this.

alphaho:
@987Nabil I had the same question and it is very likely to be a breaking change. Let me close this PR and wait for a major version upgrade.

Issue #588, opened 2023-08-04
Support validation for optional values


williamareynolds:
Example:

```scala

case class User(
    @validate(Validation.email)
    email: Option[String]
)

```

Currently, there doesn't appear to be any support for performing a validation on an optional field. My expectation of the normal behavior would be that the validation would only be applied when the value is not `None`, or that it can be controlled with combinators. I think this could be achieved via something like this:

```scala
case class User(
  @validate(Validation.optional(Validation.email))
  email: Option[String] // can be empty and still pass validation
)

PR #624, opened 2023-12-24
Added new methods for Validation

pablf:
Add methods to create Validations for `Option` and `Either` types. It also allows to create new validations with method `premap`. Adresses #588.
Use:
```
case class User(
  @validate(Validation.email.optional()) // same as @validate(Validation.email.optional(true))
  email: Option[String] // can be empty and still pass validation
)
```

```
case class User(
  @validate(Validation.email.optional(false))
  email: Option[String] // doesn't pass validation when empty
)
```

### DIFF
--- a/zio-schema/shared/src/main/scala/zio/schema/validation/Bool.scala
+++ b/zio-schema/shared/src/main/scala/zio/schema/validation/Bool.scala
@@ -5,11 +5,11 @@ sealed trait Bool[A] { self =>
   def ||(that: Bool[A]): Bool[A] = Bool.Or(self, that)
   def unary_! : Bool[A]          = Bool.Not(self)
 
-  def map[B](f: A => B, notCounter: Int = 0): Bool[B] = self match {
-    case Bool.And(left, right) => Bool.And(left.map(f, notCounter), right.map(f, notCounter))
-    case Bool.Or(left, right)  => Bool.Or(left.map(f, notCounter), right.map(f, notCounter))
+  def map[B](f: A => B): Bool[B] = self match {
+    case Bool.And(left, right) => Bool.And(left.map(f), right.map(f))
+    case Bool.Or(left, right)  => Bool.Or(left.map(f), right.map(f))
     case Bool.Leaf(value)      => Bool.Leaf(f(value))
-    case Bool.Not(value)       => Bool.Not(value.map(f, notCounter + 1))
+    case Bool.Not(value)       => Bool.Not(value.map(f))
   }
 }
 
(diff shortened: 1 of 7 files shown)


zio-schema/shared/src/main/scala/zio/schema/validation/Bool.scala
```diff
@@ -4,6 +4,13 @@ sealed trait Bool[A] { self =>
   def &&(that: Bool[A]): Bool[A] = Bool.And(self, that)
   def ||(that: Bool[A]): Bool[A] = Bool.Or(self, that)
   def unary_! : Bool[A]          = Bool.Not(self)
+
+  def map[B](f: A => B, notCounter: Int = 0): Bool[B] = self match {
```
vigoo, line 8: This `notCounter` parameter is not used for anything

zio-schema/shared/src/main/scala/zio/schema/validation/Predicate.scala
```diff
@@ -2,10 +2,11 @@ package zio.schema.validation
 
 import zio.Chunk
 
-sealed trait Predicate[A] {
+sealed trait Predicate[A] { self =>
   type Errors = Chunk[ValidationError]
   type Result = Either[Errors, Errors]
   def validate(value: A): Result
+  def premap[B](f: B => A): Predicate[B] = Predicate.Premap(Bool.Leaf(self), f)
```
vigoo, line 9: Shouldn't this called `contramap`?

zio-schema/shared/src/main/scala/zio/schema/validation/Predicate.scala
```diff
@@ -69,4 +70,27 @@ object Predicate {
...
+  final case class Optional[A](pred: Bool[Predicate[A]], validNone: Boolean) extends Predicate[Option[A]] {
+
+    def validate(value: Option[A]): Result = value match {
+      case None =>
+        if (validNone) Right(Chunk(ValidationError.EqualToNone())) else Left(Chunk(ValidationError.EqualToNone()))
+      case Some(v) => Validation(pred).validate(v).map(_ => Chunk.empty)
+    }
+  }
+
+  final case class Contramap[B, A](pred: Bool[Predicate[A]], f: (B => A)) extends Predicate[B] {
```
jdegoes, line 83: Storing a Scala function inside `Contramap` may complicate the story of converting a validation into a JSON Spec. Is contramap strictly required?

jdegoes (approved):

PR #627, opened 2023-12-26
add support for transformed primitve map keys

DannyMor:
currently, defining a schema for maps with complex keys that represent simple keys is impossible. For example, consider a map defined as `Map[Key, Value]`, where `Key` is defined as `case class Key(key:String)`. When you try to define a schema like this:

```scala
schema = Schema.map(
   Schema[String].transform(Key.apply, _.key), 
   ...
)
```
to decode data that looks like this:

```json
{
     "some_key": { 
          "value": "some_value" 
     },
     "some_other_key": { 
          "value": "some_other_value"
     }
}
```

it will fail and produce an error: `"expected '[' got '{'"`. Currently, any non-primitive key (or lazy) defaults to an array representation of a key-value pair, causing the issue.

To address this issue, this PR proposes that `jsonFieldEncoder` and `jsonFieldDecoder` should also take into account the `Transform` schema type. This means it will recursively check if the internal schema is indeed a primitive codec, ensuring that the data is parsed correctly from a simple key, and applying the transformation from the key schema definition accordingly.

### DIFF
--- a/zio-schema-json/shared/src/main/scala/zio/schema/codec/JsonCodec.scala
+++ b/zio-schema-json/shared/src/main/scala/zio/schema/codec/JsonCodec.scala
@@ -283,8 +283,12 @@ object JsonCodec {
           else
             throw new Exception(s"Missing a handler for encoding of schema ${schema.toString()}.")
       }
+    //scalafmt: { maxColumn = 120, optIn.configStyleArguments = true }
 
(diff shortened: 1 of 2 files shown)


vigoo (approved):
Thank you, it looks like a much better default behavior. I was wondering if we should introduce some kind of annotation that forces the old behavior to help migration for people who already used the library for encoding maps with transformed keys - but probably that would have been reported as an issue previously so let's just do this and create a fallback to the old behavior only if anyone needs it.

PR #636, opened 2024-01-13
Support sealed abstract class in DeriveSchema

### LINKED ISSUE #621: Deriver macro is not able to support sealed abstract classes
I noticed that recently a Schema was implemented for the zio-json Json AST, which is defined as a sealed abstract class. The relevant code can be found here: [zio-schema package.scala L13](https://github.com/zio/zio-schema/blob/main/zio-schema-json/shared/src/main/scala/zio/schema/codec/package.scala#L13).

I am wondering if it would be feasible to introduce functionality for generating derivers for sealed abstract classes. This feature would be very beneficial for some of the work in zio-jdbc towards supporting JSON column types. While a general implementation would be ideal, even a special-case solution for Json AST's would be greatly appreciated.

Looking forward to your thoughts on this.

vigoo:
Resolves #621 
/claim #621 

I don't fully understand the original issue (about "only supporting JSON" and how is it related to zio-jdbc) but this PR adds support for `DeriveSchema` for `sealed abstract class` types, treating them as enums, supporting an arbitrary level of abstract classes if the leaf classes are case classes.

So for example these two now works:

```scalas
  sealed abstract class AbstractBaseClass(val x: Int)
  final case class ConcreteClass1(override val x: Int, y: Int)    extends AbstractBaseClass(x)
  final case class ConcreteClass2(override val x: Int, s: String) extends AbstractBaseClass(x)

  sealed abstract class AbstractBaseClass2(val x: Int)
  sealed abstract class MiddleClass(override val x: Int, val y: Int)                   extends AbstractBaseClass2(x)
  final case class ConcreteClass3(override val x: Int, override val y: Int, s: String) extends MiddleClass(x, y)
```

the first generates an `Enum2` with with two cases containing two records,  the second a nested `Enum1` within an `Enum1` leading to the leaf record.

Note that the Scala 3 version of `DeriveSchema` already supported this.


987Nabil:
The only connection to json is, that the json is a abstract class. I think you can ignore that information

zio-schema-derivation/shared/src/main/scala-2/zio/schema/DeriveSchema.scala
```diff
@@ -65,11 +68,13 @@ object DeriveSchema {
       } else if (isCaseClass(tpe)) deriveRecord(tpe, stack)
       else if (isSealedTrait(tpe))
         deriveEnum(tpe, stack)
+      else if (isSealedAbstractClass(tpe))
```
987Nabil, line 71: Is there a need to make a difference here? How about just `isSealed` and join the two cases

vigoo: No it turned out to be the same. I think keeping the two explicit `isXXX` functions is good for understandability, but we can merge the two branches

jdegoes (approved):

PR #637, opened 2024-01-14
Support for explicit field numbers and skipping unknown fields during deserialization

### LINKED ISSUE #575: Support for custom and reserved field numbers in Protobuf codec
When schema evolves and a field is deleted, it is common to use `reserved n;` (with `n` being the deleted field number) in proto files to make sure that a given field number is not reused (which would be incompatible with existing data that has this field). It might also happen that you want a field to have a given field number for backward compatibility reasons.

A way to support that with the Protobuf codec would be great. Maybe using annotations?

vigoo:
Resolves #575 
/claim #575


zio-schema-protobuf/shared/src/main/scala/zio/schema/codec/FieldMappingCache.scala
```diff
@@ -0,0 +1,37 @@
...
+
+import scala.collection.mutable
+
+import zio.schema.Schema
+
+/**
+ * A per-encooding/decoding cache for field mappings. No need for thread safety as a single encoding/decoding
+ * is sequential.
+ */
+class FieldMappingCache {
```
jdegoes, line 11: Maybe worth making the constructor private so you have to go through `FieldMapping`?

vigoo: Too much Rust coding, I though it's private because there is no `pub` :D

jdegoes (approved):

Issue #651, opened 2024-01-30
Scala 3: Enum case annotations are the parent annotations for derived schema


987Nabil:
If given an enum with annotations

```scala3
@description("Color Enum")
enum Color:
  @description("Red")
  case Red
```

The `@description("Color Enum")` annotation will be added to the schema derived via `DeriveSchema[Color]` as an annotation to the top level schema, the case `Red` and the schema of the `Red` case. While `@description("Red")` is nowhere to be found.

This is clearly a bug. I think the annotations for the case and the case schema should be the ones added to `case Red`

PR #657, opened 2024-01-31
Correctly derive annotations for enum cases

pablf:
/claim #651


987Nabil:
@pablf Could you uncomment the test assertion in line 90, please? That was the assertion that made me discover the bug.

987Nabil (approved):

PR #670, opened 2024-03-25
Improve orElse, to auto widen for sum types and remove ClassTags

987Nabil:
@jdegoes I realised, that for

```scala
enum Foo:
  case A, B
```

The current version would return `Schema[A | B]` for `Schema[A].orElse(Schema[B])` but with this change it would return `Schema[Foo]`.
Also the inline removes the need for class tags


vigoo:
Does it still infer `A | B` if they are independent from each other, and not `Any`?

987Nabil:
Yes, it does. This is the same technique used by `Either#merge` which works also well with Scala 3 union types.

vigoo (approved):

PR #672, opened 2024-03-26
Option json decoder hands over type instead of option parsing errors

987Nabil:


987Nabil:
I actually think, this fixes also some bug. I build this locally and instead of seeing the error I wanted to expose, it just works.
Therefore I'll merge this so I can work with the snapshot

PR #677, opened 2024-04-17
Fix the json decoder for case classes that only have `None` fields.

### LINKED ISSUE #661: Error decoding empty object with optional fields
I believe there is currently a bug when trying to decode an empty object `{}` when the schema is for a case class with all `Optional` fields resulting in a `expected '"' got '}'` error. 

## Evidence:
Upgrading `zio-openai` from `0.4.0` to `0.4.1` resulted in a json decoding error for responses which decoded fine in the prior version (giving `Encountered openai error: Unknown(zio.schema.codec.DecodeError$ReadError: .choices[0].delta(expected '"' got '}'))` error on the final event of a chat response SSE stream). The json it was trying to decode was:
```
{"id":"chatcmpl-8qGJgtg8np7wvJuh1SfJRGY3SzLXn","object":"chat.completion.chunk","created":1707466468,"model":"gpt-3.5-turbo-0613","system_fingerprint":null,"choices":[{"index":0,"delta":{},"logprobs":null,"finish_reason":"stop"}]}
```
With relevant part of schema:
```
final case class ChoicesItem(delta: ChatCompletionStreamResponseDelta, logprobs: Optional[CreateChatCompletionStreamResponse.ChoicesItem.Logprobs] = ???, finishReason: Optional[FinishReason], index: Int)

final case class ChatCompletionStreamResponseDelta(content: Optional[String] = ???, functionCall: Optional[ChatCompletionStreamResponseDelta.FunctionCall] = ???, toolCalls: Optional[Chunk[ChatCompletionMessageToolCallChunk]] = ???, role: Optional[ChatCompletionStreamResponseDelta.Role] = ???)
```

Further investigation determined that pinning `dev.zio::zio-schema-json:0.4.17` to `dev.zio::zio-schema-json:0.4.16` fixed the decoding issue. The commits going from 0.4.16 -> 0.4.17 can be found [here](https://github.com/zio/zio-schema/compare/v0.4.16...v0.4.17#diff-2b3004efa6280843eec4f1cc41c98f89b34df187c716a64fa73784fe78cbc9e5L1154). 

Looking at those commits, I think the issue come from this [line](https://github.com/runtologist/zio-schema/blob/64fe606968b11b6b8a85a7c3c5086922dba4df97/zio-schema-json/shared/src/main/scala/zio/schema/codec/JsonCodec.scala#L1263) in #619.

## Possible solution:
Current code in `unsafeDecodeFields`:
```
@tailrec
      def loop(index: Int, in: RetractReader): Unit = {
        val fieldRaw = Lexer.field(trace, in, new StringMatrix(aliasesMatrix))
        if (index == discriminator) {
          Lexer.skipValue(trace, in)
        } else {
          fieldRaw match {
            case -1 if index == discriminator => Lexer.skipValue(trace, in)
            case -1 if rejectExtraFields      => throw UnsafeJson(JsonError.Message("extra field") :: trace)
            case -1                           => Lexer.skipValue(trace, in)
            case idx =>
              val field = fieldAliases.getOrElse(aliasesMatrix(idx), -1)
              if (buffer(field) != null)
                throw UnsafeJson(JsonError.Message("duplicate") :: trace)
              else
                buffer(field) = schemaDecoder(schemas(field)).unsafeDecode(spans(field) :: trace, in)
          }
        }
        if (Lexer.nextField(trace, in)) loop(index + 1, in)
      }

      if (discriminator == -1) {
        Lexer.char(trace, in, '{')
        loop(0, in)
      } else if (discriminator == -2) {
        if (Lexer.nextField(trace, in)) loop(0, in)
      } else {
        val rr = RecordingReader(in)
        if (Lexer.firstField(trace, rr)) loop(0, rr)
      }
```
If the discriminator is -1 (meaning no discriminator), then the Lexer reads in the { and immediately starts the loop. The loop calls Lexer.field(trace, in, new StringMatrix(aliasesMatrix))  which immediately tries reading in a field (starting with " ) and throws an error if it doesnt see it. I think what we need to do is change the code to: 
```
if (discriminator == -1) {
    Lexer.char(trace, in, '{')
    if(Lexer.firstField(trace, in)) loop(0, in)
}
```

paulpdaniels:
Fixes #661

### DIFF
--- a/zio-schema-json/shared/src/main/scala/zio/schema/codec/JsonCodec.scala
+++ b/zio-schema-json/shared/src/main/scala/zio/schema/codec/JsonCodec.scala
@@ -1341,7 +1341,7 @@ object JsonCodec {
 
       if (discriminator == -1) {
         Lexer.char(trace, in, '{')
-        loop(0, in)
+        if (Lexer.firstField(trace, in)) loop(0, in)
       } else if (discriminator == -2) {
         if (Lexer.nextField(trace, in)) loop(0, in)
       } else {
--- a/zio-schema-json/shared/src/test/scala-2/zio/schema/codec/JsonCodecSpec.scala
+++ b/zio-schema-json/shared/src/test/scala-2/zio/schema/codec/JsonCodecSpec.scala
@@ -1096,6 +1096,9 @@ object JsonCodecSpec extends ZIOSpecDefault {
       },
       test("object") {
         assertEncodesThenDecodes(schemaObject, Singleton)
+      },
+      test("all optional fields empty") {
+        assertEncodesThenDecodes(AllOptionalFields.schema, AllOptionalFields(None, None, None))
       }
     ),
     suite("record")(
@@ -1802,5 +1805,15 @@ object JsonCodecSpec extends ZIOSpecDefault {
   object MapOfComplexKeysAndValues {
     implicit lazy val mapSchema: Schema[Map[KeyWrapper, ValueWrapper]] = Schema.map[KeyWrapper, ValueWrapper]
     implicit lazy val schema: Schema[MapOfComplexKeysAndValues]        = DeriveSchema.gen[MapOfComplexKeysAndValues]
+  }
+
+  final case class AllOptionalFields(
+    name: Option[String],
+    mode: Option[Int],
+    active: Option[Boolean]
+  )
+
+  object AllOptionalFields {
+    implicit lazy val schema: Schema[AllOptionalFields] = DeriveSchema.gen[AllOptionalFields]
   }
 }


987Nabil (approved):

jdegoes (approved):

paulpdaniels:
I’d love to merge it, though I’m not clear on the broken website build. @987Nabil do you know what I need to do there?

jdegoes:
cc @khajavi

khajavi:
@jdegoes 
This is because the `zio-sbt-ci` plugin used for zio-schema is outdated! and I mentioned it [here](https://github.com/zio/zio-schema/issues/679) with a detailed solution. Anyway, I'll take this tomorrow to fix that.

987Nabil:
@paulpdaniels just sbt docs/generateReadme commit and push

PR #688, opened 2024-06-09
fix: handle missing collection fields in JSON decoding

### LINKED ISSUE #664: Json Codec: Read missing field as empty collection
Currently, if a field that is a collection is missing, that will lead to a decoding error.

Either by default or configurable, if a field that is a collection is missing, we should just return an empty collection.

onyedikachi-david:
fixes #664
/closes #664
/claim #664

### DIFF
--- a/zio-schema-json/shared/src/main/scala/zio/schema/codec/JsonCodec.scala
+++ b/zio-schema-json/shared/src/main/scala/zio/schema/codec/JsonCodec.scala
@@ -66,10 +66,10 @@ object JsonCodec {
         )
     }
 
-  implicit def schemaBasedBinaryCodec[A](implicit schema: Schema[A]): BinaryCodec[A] =
+  implicit def schemaBasedBinaryCodec[A](implicit schema: Schema[A], config: JsonCodec.Config): BinaryCodec[A] =
     schemaBasedBinaryCodec[A](JsonCodec.Config.default)
 
-  def schemaBasedBinaryCodec[A](cfg: Config)(implicit schema: Schema[A]): BinaryCodec[A] =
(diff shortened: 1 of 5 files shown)


zio-schema-json/shared/src/main/scala/zio/schema/codec/JsonCodec.scala
```diff
@@ -1355,17 +1362,37 @@ object JsonCodec {
...
+    private def isCollectionSchema(schema: Schema[_]): Boolean =
+      schema match {
+        case Schema.Sequence(_, _, _, _, _) => true
+        case Schema.Set(_, _)               => true
+        case Schema.Map(_, _, _)            => true
+        case _                              => false
+      }
+
+// Helper method to create an empty collection based on the schema
+    private def createEmptyCollection(schema: Schema[_]): Any =
```
987Nabil, line 1388: I think this logic belongs to the `trait Collection` in `Schema.scala`, so each collection schema has to implement a method that creates an empty collection.

zio-schema-json/shared/src/main/scala/zio/schema/codec/JsonCodec.scala
```diff
@@ -104,13 +104,13 @@ object JsonCodec {
   def jsonEncoder[A](cfg: JsonCodec.Config)(schema: Schema[A]): ZJsonEncoder[A] =
     JsonEncoder.schemaEncoder(schema, cfg)
 
-  def jsonDecoder[A](schema: Schema[A]): ZJsonDecoder[A] =
+  def jsonDecoder[A](schema: Schema[A])(implicit config: JsonCodec.Config): ZJsonDecoder[A] =
```
987Nabil, line 107: Changes are binary incompatible.
Config should not be implicit.
You should not reuse the same config value, since these are two different features. They should be configurable independent of each other.

987Nabil (changes requested):

PR #692, opened 2024-06-14
Fix code OpenAPI documentation generator crashes on parameterized case

Anshgrover23:
/claim #2767
fixes: #2767


987Nabil:
Is not a fix. Best case a work around. But this code does not belong in zio-schema

PR #703, opened 2024-06-29
Save info about generically applied types in new annotation (#695)

### LINKED ISSUE #695: Add infos about generic parameters and what fields are of such a generic type to schemas
See https://github.com/zio/zio-http/issues/2767

987Nabil:
fixes #695
/claim #695


987Nabil:
@jdegoes I used an annotation, since it is easy to implement and even more important non-breaking.

jdegoes (approved):

PR #704, opened 2024-06-29
JsonCodec respects annotations for GenericRecord (#700)

### LINKED ISSUE #700: JsonDecoder should take into account all annotations on `EnumN` and `GenericRecord`
The `enumEncoder/Decoder` and the `recordEncoder/Decoder` almost ignore all annotations on fields. Something like mapped names with the `@fieldName` annotation are working, because they are picked up by the macro and set as name of the fields. But we should not rely on it. If I create a Schema by hand and add the annotation or add the annotation after the macro, it should still work.

### LINKED ISSUE #701: `@fieldName` annotation breaks JsonCodec for 22+ field case classes
Adding the `@fieldName` annotation to the case class for the GenericRecord test in `JsonCodecSpec`

```scala
 case class RecordExample(
    @fieldName("$f1") f1: Option[String]
```

Leads to
```
Exception in thread "zio-fiber-914" zio.schema.codec.DecodeError$ReadError: (Field f1 is missing)
```

987Nabil:
fixes #700
fixes #701
/claim #700
/claim #701


987Nabil:
The enum part is still missing

jdegoes (approved):

PR #706, opened 2024-06-30
Check type before using `asInstanceOf`

### LINKED ISSUE #668: JSON codec built from auto-derived schema fails for enumeration with intermediate type
```scala
//> using scala 3.4.0
//> using dep dev.zio::zio-schema:1.0.1
//> using dep dev.zio::zio-schema-derivation:1.0.1
//> using dep dev.zio::zio-schema-json:1.0.1

sealed trait Animal
object Animal:
  sealed trait Mammal extends Animal
  case object Bison extends Mammal
  given schema: zio.schema.Schema[Animal] = zio.schema.DeriveSchema.gen[Animal]
  given jsonCodec: zio.json.JsonCodec[Animal] = zio.schema.codec.JsonCodec.jsonCodec(schema)

  @main def main: String =
    import zio.json.EncoderOps
    val value: Animal = Bison
    // Exception in thread "main" java.lang.ClassCastException:
    //   class zio.schema.Schema$Enum1 cannot be cast to class zio.schema.Schema$CaseClass0
    value.toJson
```

pablf:
fixes #668
/claim #668

### DIFF
--- a/zio-schema-json/shared/src/main/scala/zio/schema/codec/JsonCodec.scala
+++ b/zio-schema-json/shared/src/main/scala/zio/schema/codec/JsonCodec.scala
@@ -378,7 +378,9 @@ object JsonCodec {
 
     private def enumEncoder[Z](schema: Schema.Enum[Z], cfg: Config, cases: Schema.Case[Z, _]*): ZJsonEncoder[Z] =
       // if all cases are CaseClass0, encode as a String
-      if (schema.annotations.exists(_.isInstanceOf[simpleEnum])) {
+      if (schema.annotations.exists(_.isInstanceOf[simpleEnum]) && cases.forall(
+            _.schema.isInstanceOf[Schema.CaseClass0[_]]
+          )) {
         val caseMap: Map[Z, String] =
           schema.nonTransientCases
             .map(
@@ -699,7 +701,9 @@ object JsonCodec {
       }.toMap
 
       // if all cases are CaseClass0, decode as String
-      if (cases.forall(_.schema.isInstanceOf[Schema.CaseClass0[_]])) {
+      if (parentSchema.annotations.exists(_.isInstanceOf[simpleEnum]) && cases.forall(
+            _.schema.isInstanceOf[Schema.CaseClass0[_]]
+          )) {
         val caseMap: Map[String, Z] =
           cases.map(case_ => case_.id -> case_.schema.asInstanceOf[Schema.CaseClass0[Z]].defaultConstruct()).toMap
         ZJsonDecoder.string.mapOrFail(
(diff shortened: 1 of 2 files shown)


987Nabil:
@pablf can you please add fixes #XXX when you open PRs, so that the issues get auto closed when the PR is merged? Thanks :)

zio-schema-json/shared/src/main/scala/zio/schema/codec/JsonCodec.scala
```diff
@@ -378,7 +378,9 @@ object JsonCodec {
 
     private def enumEncoder[Z](schema: Schema.Enum[Z], cfg: Config, cases: Schema.Case[Z, _]*): ZJsonEncoder[Z] =
       // if all cases are CaseClass0, encode as a String
-      if (schema.annotations.exists(_.isInstanceOf[simpleEnum])) {
+      if (schema.annotations.exists(_.isInstanceOf[simpleEnum]) && cases.forall(
```
987Nabil, line 381: sine `@simpleEnum` is not added explicitly by the user, is this not an error in the macro?
And if the user would add it, the macro should fail.
But here, the user could have created a schema manually or added the annotation manually after the macro code.
So I think we should at least print a warning here. I would even prefer to fail.

987Nabil (changes requested):

zio-schema-json/shared/src/main/scala/zio/schema/codec/JsonCodec.scala
987Nabil: @pablf I think it is okay, if all leaves are objects or case classes without fields, that it is a simple enum. But why would you need the check in the json codec for that? The check there makes sense to prevent errors in case of an annotation where no annotation belongs, to raise an error. But if leaves are simple cases and the json codec still can't handle it, it is a different error and this would not be the fix.
I think the goals should be
- simple leaves = simple enum
- simple leaves with intermediate traits should be handled by codecs as if the intermediate trait is missing
- no simple leaves but annotated => macro fails
- no simple leaves but annotated => codec fails

987Nabil: Maybe we need more issues

987Nabil: I think it should not. At least I do not see any value in it. @jdegoes wdyt?

jdegoes: @987Nabil Why not derive intermediate traits??

987Nabil: I don't say it's not good. I say, is there a use case for it? Is there value in it? Or does it just make the use case of simple enum harder without a clear benefit.

Anyhow, if we have intermediate traits derived, it should be the same for Scala 2 and 3 and leave traits (no subtypes) should be ignored for the question of "Is this a simple enum".

jdegoes: It seems like not deriving intermediates makes the system "lossy", which is something I like to avoid on principle even if I don't have a compelling use case for it.

987Nabil: I am fine with that. Then the Scala 2 macro should do the same. And the structure should be interpreted correctly regarding simple enum

PR #710, opened 2024-07-04
Fix JSON decoding of empty objects

guersam:
This PR fixes the decoding failure of discriminated case objects.

### Example

```scala
@discriminatorName("type")
enum Foo derives Schema:
  case Bar(b: Int)
  case Baz

JsonCodec.jsonDecoder(Schema[Foo]).decodeJson("""{ "type": "Baz" }""")
```

It also makes case object decoding respect `@rejectExtraFields`. 

However, currently `@rejectExtraFields` can be applied only to the leaf nodes of ADTs, not the sealed trait/enum itself. To fix this, we need to propagate the parent schema to every caseClassN decoder, but it's out of the scope of this PR. (#700)

### DIFF
--- a/zio-schema-json/shared/src/main/scala/zio/schema/codec/JsonCodec.scala
+++ b/zio-schema-json/shared/src/main/scala/zio/schema/codec/JsonCodec.scala
@@ -991,7 +991,29 @@ object JsonCodec {
     import JsonCodec.JsonDecoder.schemaDecoder
 
     private[codec] def caseClass0Decoder[Z](discriminator: Int, schema: Schema.CaseClass0[Z]): ZJsonDecoder[Z] = { (trace: List[JsonError], in: RetractReader) =>
-      if (discriminator == -1) Codecs.unitDecoder.unsafeDecode(trace, in)
+      def skipField(): Unit = {
+        val rejectExtraFields = schema.annotations.collectFirst({ case _: rejectExtraFields => () }).isDefined
+        if (rejectExtraFields) {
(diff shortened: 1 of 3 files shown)


zio-schema-json/shared/src/main/scala/zio/schema/codec/JsonCodec.scala
```diff
@@ -991,7 +991,29 @@ object JsonCodec {
     import JsonCodec.JsonDecoder.schemaDecoder
 
     private[codec] def caseClass0Decoder[Z](discriminator: Int, schema: Schema.CaseClass0[Z]): ZJsonDecoder[Z] = { (trace: List[JsonError], in: RetractReader) =>
-      if (discriminator == -1) Codecs.unitDecoder.unsafeDecode(trace, in)
+      def skipField(): Unit = {
+        val rejectExtraFields = schema.annotations.collectFirst({ case _: rejectExtraFields => () }).isDefined
```
guersam, line 995: Calculating this every time may seem redundant, but `skipField()` won't run at all in most cases.

987Nabil: I still think we can make this a val on `Schema.Record`

jdegoes (approved):

guersam:
/claim #711 

@jdegoes Can I claim a bounty with a PR merged in advance?

987Nabil:
@guersam ask algora ppl in the discord

PR #718, opened 2024-08-04
Remove unused `recordName` annotation

### LINKED ISSUE #705: The `@recordName` annotation is unsued
There is an annotation `recordName` that is currently unused.
We should either use it or remove it.
```scala
/**
 * Annotation for specifying an alternate name for a record.
 *
 * @param name The alternate name to use for the record.
 */
final case class recordName(name: String) extends StaticAnnotation
```

Neptune650:
Fixes #705 
/claim #705 

Let me know if there's anything else to be done, from my research, this is the only codec with this.


987Nabil:
does not do what was asked for

harshsbhat:
Hey, @987Nabil  can you please tell me what you are expecting here? I though this PR correctly removes the recordName annotation which I unused

Neptune650:
> does not do what was asked for

What's being asked for then

PR #721, opened 2024-08-05
Run all possible test for all Scala 3 (#709)

### LINKED ISSUE #709: Ensure non Scala version specific specs being executed for Scala 3 as well
A lot of specs are in the `scala-2` source folders. For example the `JsonCodecSpec`.
This prevents them from being executed on for Scala 3.
After taking a short look, it seems this will take some time to migrate, since the tests depend on some generators with complex types. These types are evaluated more strict or just different by the Scala 3 compiler and need to be adjusted to make it work across Scala 2 and 3.

@jdegoes I think this might be also a reason why we had some bugs reports, where zio-schema was acting different for Scala 3.

987Nabil:
fixes #709
/claim #709


987Nabil:
@jdegoes the native build is flakey. I already opened an issue for it

jdegoes (approved):

PR #723, opened 2024-08-07
Non empty collection schemas (#717)

### LINKED ISSUE #717: Schema for non empty collections
We should support some kind of non empty collections. 
Either
1. we create a new sub type of `Collection` that takes care of the validation on construction
2. we have a method `def nonEmptyChunk[A :Schema]: Schema[NonEmptyChunk[A]]` that uses transform to validate the non-emptiness on construction

987Nabil:
@jdegoes I went with a new Schema sub type instead of transforming, since it offers a way to introspect in libs. For example, in zio-http query parsing, we can just check that the query params is not there and fail immediately.
Also it is easier to add support for other non empty collections by the user.

fixes #717
/claim #717


jdegoes (dismissed):

jdegoes:
@987Nabil Good to merge when CI is passing. 👍

PR #727, opened 2024-08-14
Removed the all the usage of recordName

harshsbhat:
The #705 asks to remove ALL the instances of the recordName annotations as it is unused. 

fixes: #705 
/claim #705

If you want to USE it specifically instead of removing it. Please specify that while closing the PR


jdegoes:
I think the $250 bounty I was thinking it would be _used_ in all the codecs, not deleted. The codecs are just ignoring it right now.

PR #735, opened 2024-08-31
feat: upgrade scala native to latest and enable multithreading

### LINKED ISSUE #720: Fix Scala native pipeline
The Scala Native build fails some times. Maybe a OOM issue. An upgrade to Scala native 0.5 might help.

asr2003:
Upgraded ZIO and Scala-Native to their latest versions and while enabling `multithreading=true`, it resulted in the Out of heap space which is an issue with scala-native and an active discussion is going here in the upstream:

- https://github.com/scala-native/scala-native/issues/4032

Note that the GC change is only needed for few tests which are memory intensive. Once that's upstream is fixed I think we can remove that config safely

fixes #720
/claim #720

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,9 +49,9 @@ jobs:
         jvm: ${{ matrix.java }}
     - name: Cache scala dependencies
       uses: coursier/cache-action@v6
-    - name: Install libuv
+    - name: Install Bohem GC
       if: matrix.platform == 'Native'
-      run: sudo apt-get update && sudo apt-get install -y libuv1-dev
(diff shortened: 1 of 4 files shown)


asr2003:
@987Nabil I didn't get it what's wrong with my PR 🙄 and closed

I have seen yesterday the issue and I am also one of who tried  enabling
multithreading in Zio along with @varshith257,  so I have good context of
this upgradation and Multithreading issue, so seems the same work of Zio
here and raised PR.

I

asr2003:
> zio/scala-native is something we want

Yes I have done the same of upgrading zio and native .

987Nabil:
@asr2003 #720 mentions clearly fixing the CI. The CI was already fixed in #726

PR #738, opened 2024-09-09
Fixes #737

### LINKED ISSUE #737: zio.schema.codec.JsonCodec should be able to decode multiple values from a stream and separate encoded values into a stream
Currently `JsonCodec` only decodes the first element from a stream, because it first concatenates the whole stream and then parses it, ignoring everything which comes after the first element.

When elements are encoded into the stream, they are concatenated without any separator. Such streams then cannot be decoded, if the element type is numeric.

Propsal: 
* When encoding streams, separate the the elements with a new line.
* When decoding streams, parse individual elements on the JSON boundaries.

gregor-rayman:
Fix for the issue #737:

* separates encoded elements with new lines (so that numbers can be decoded separated)
* when decoding, looks for boundaries between elements. Needs separators only for numbers. Recognizes automatically the ends of objects, arrays, strings, boolean values and nulls.

### DIFF
--- a/zio-schema-json/shared/src/main/scala/zio/schema/codec/JsonCodec.scala
+++ b/zio-schema-json/shared/src/main/scala/zio/schema/codec/JsonCodec.scala
@@ -137,7 +137,7 @@ object JsonCodec {
 
           if (valueEnded && depth == 0) {
             val str = stringBuilder.result()
-            if (!str.isBlank) chunkBuilder += str
+            if (!str.forall(_.isWhitespace)) chunkBuilder += str
             stringBuilder.clear()
           }
         }
--- a/zio-schema-json/shared/src/main/scala/zio/schema/codec/JsonCodec.scala
+++ b/zio-schema-json/shared/src/main/scala/zio/schema/codec/JsonCodec.scala
@@ -65,11 +65,19 @@ object JsonCodec {
     schemaBasedBinaryCodec[A](JsonCodec.Config.default)
 
   val splitOnJsonBoundary: ZPipeline[Any, Nothing, String, String] = {
-    val validNumChars = Set('0', '1', '2', '3', '4', '5', '6', '7', '8', '9', 'E', 'e', '-', '+', '.')
+    val validNumChars          = Set('0', '1', '2', '3', '4', '5', '6', '7', '8', '9', 'E', 'e', '-', '+', '.')
+    val ContextJson            = 'j'
+    val ContextString          = 's'
+    val ContextBoolean         = 'b'
+    val ContextNull            = 'u'
+    val ContextNullAfterFirstL = 'x'
+    val ContextNumber          = 'n'
+    val ContextEscape          = 'e'
+
     ZPipeline.suspend {
       val stringBuilder = new StringBuilder
       var depth         = 0
-      var valueType     = 'j' // j = json, s = string, b = boolean, u = null, n = number, x = null after first null, e = escape in string
+      var context       = ContextJson
 
       def fetchChunk(chunk: Chunk[String]): Chunk[String] = {
         val chunkBuilder = ChunkBuilder.make[String]()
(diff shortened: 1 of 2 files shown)


zio-schema-json/shared/src/test/scala/zio/schema/codec/JsonCodecSpec.scala
```diff
@@ -436,6 +437,71 @@ object JsonCodecSpec extends ZIOSpecDefault {
           """{"NumberOne":{"value":"foo"}}"""
         )
       }
+    ),
+    suite("Streams")(
+      test("Encodes a stream with multiple integers") {
+        assertEncodesMore(Schema[Int], 1 to 5, charSequenceToByteChunk("1\n2\n3\n4\n5"))
+      },
+      test("Decodes a stream with multiple integers") {
```
987Nabil, line 447: I would make this a suite and the different checks tests. Easier to read and tests names can give intention

zio-schema-json/shared/src/main/scala/zio/schema/codec/JsonCodec.scala
```diff
@@ -62,14 +58,110 @@ object JsonCodec {
...
 
   implicit def schemaBasedBinaryCodec[A](implicit schema: Schema[A]): BinaryCodec[A] =
     schemaBasedBinaryCodec[A](JsonCodec.Config.default)
 
+  val splitOnJsonBoundary: ZPipeline[Any, Nothing, String, String] = {
+    val validNumChars = Set('0', '1', '2', '3', '4', '5', '6', '7', '8', '9', 'E', 'e', '-', '+', '.')
+    ZPipeline.suspend {
+      val stringBuilder = new StringBuilder
+      var depth         = 0
+      var valueType     = 'j' // j = json, s = string, b = boolean, u = null, n = number, x = null after first null, e = escape in string
```
987Nabil, line 72: Should we make val with this values? I don't thinkit would be a significant performance issue, but might be easier to understand than this

987Nabil (commented):
Minor things. Looks good.

zio-schema-json/shared/src/main/scala/zio/schema/codec/JsonCodec.scala
gregor-rayman: I was inspired by `ZPipeline.splitLines` that also uses vars. The `valueType` is used for context to recognize how to interpret the current character and to spot when a section ends.

zio-schema-json/shared/src/test/scala/zio/schema/codec/JsonCodecSpec.scala
gregor-rayman: Good point, I'll refactor

zio-schema-json/shared/src/main/scala/zio/schema/codec/JsonCodec.scala
gregor-rayman: Or did you mean using an enum / sealed trait?

987Nabil: No just vals. Like a constants for the different states. Pretty sure a sealed trait is less performant

zio-schema-json/shared/src/test/scala/zio/schema/codec/JsonCodecSpec.scala
```diff
@@ -1505,6 +1571,23 @@ object JsonCodecSpec extends ZIOSpecDefault {
     assertZIO(stream)(equalTo(chunk))
   }
 
+  private def assertEncodesMore[A](
```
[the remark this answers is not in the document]
gregor-rayman: Renamed

zio-schema-json/shared/src/test/scala/zio/schema/codec/JsonCodecSpec.scala
gregor-rayman: Refactored to suites

zio-schema-json/shared/src/main/scala/zio/schema/codec/JsonCodec.scala
gregor-rayman: Ok, I'll do that

gregor-rayman: Done

987Nabil (approved):

PR #751, opened 2024-10-15
Use unroll plugin to fix bin compat issue

987Nabil:


sideeffffect (approved):
This looks good as far as I can tell.

PR #773, opened 2025-01-09
Fix support of fieldName annotations

plokhotnyuk:
* Fix using of fieldName annotations to rename fields (not aliasing)

* Fix ignoring non-first fieldName annotations at the field declarations

### DIFF
--- a/zio-schema-derivation/shared/src/main/scala-3/zio/schema/DeriveSchema.scala
+++ b/zio-schema-derivation/shared/src/main/scala-3/zio/schema/DeriveSchema.scala
@@ -256,7 +256,7 @@ private case class DeriveSchema()(using val ctx: Quotes) {
                tpe.asType
            }
            val annotations = paramAnns.getOrElse(label, List.empty)
-           val nameExpr = annotations.collectFirst {
+           val nameExpr = annotations.reverse.collectFirst {
              case ann if ann.isExprOf[fieldName] =>
                val fieldNameAnn = ann.asExprOf[fieldName]
                  '{${fieldNameAnn}.name}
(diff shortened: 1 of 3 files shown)


zio-schema-derivation/shared/src/main/scala-3/zio/schema/DeriveSchema.scala
```diff
@@ -256,7 +256,7 @@ private case class DeriveSchema()(using val ctx: Quotes) {
                tpe.asType
            }
            val annotations = paramAnns.getOrElse(label, List.empty)
-           val nameExpr = annotations.collectFirst {
```
987Nabil, line 259: I actually think more the one annotation of the same kind is using the lib wrong. I would keep the old impl, since it is less complicated. You annotate twice? Your problem. Just don't.

zio-schema/shared/src/main/scala/zio/schema/Schema.scala
```diff
@@ -432,16 +432,18 @@ object Schema extends SchemaPlatformSpecific with SchemaEquality {
...
-        case aliases: fieldNameAliases => aliases.aliases
-        case f: fieldName              => Seq(f.name)
-      }.flatten.toSet + name
-
     val fieldName: String = annotations.collectFirst {
       case f: fieldName => f.name
     }.getOrElse(name)
 
+    val nameAndAliases: scala.collection.immutable.Set[String] =
+      annotations.foldLeft(scala.collection.immutable.Set(fieldName)) { (acc, annotation) =>
```
987Nabil, line 440: is there a difference besides making it a fold?

zio-schema-derivation/shared/src/main/scala-3/zio/schema/DeriveSchema.scala
plokhotnyuk: The proposed solution is a workaround for reverse order of annotations in Scala 3 macros.

The better solution would be throwing a compile-time error during derivation or/and runtime error when calculating `fieldName` value in `Schema.Field` instances

zio-schema/shared/src/main/scala/zio/schema/Schema.scala
plokhotnyuk: It avoids reusing of an old (renamed) field name as an alias

987Nabil (approved):

PR #785, opened 2025-01-14
Add support of sum-type hierarchies

plokhotnyuk:
Most Scala's JSON parsers support sum-type hierarchies (including zio-json), but this test currently fails because `zio-schema-json` decoder cannot parse JSON properly serialized by encoder.


plokhotnyuk:
Moved to https://github.com/zio/zio-schema/pull/822

PR #804, opened 2025-02-08
XML codec implemented 

### LINKED ISSUE #754: XML Codec

harshtech123:
This XML codec uses Scala-xml as dependency

this codec insures fully round trip , provide ability to encode and decode xml and supports every primitive and dynamic types 

/claim #754
/closes #754


harshtech123:
@jdegoes  hii good day ! created this xml codec using scala xml liberary but i have few doubts -

1. can we use zio-xml instead of scala-xml since scala xml is only providing support for jvm and scala.js 
2. i am going to write tests as well( actualy i already written in my local ) but wants this confirmation to proceed based on that .

in scala-xml we might need to write different file for jsplatform since . The core issue is that scala-xml (as provided on the JVM) depends on Java APIs (such as SAX, e.g. org.xml.sax.InputSource, XMLReader, etc.) that simply do not exist on Scala.js. 

so i need confirmation about the liberary can we use zio-xml ( it supports JSPlatform, JVMPlatform, NativePlatform) ? or we have to go with scala-xml with jvm and js only while writing separate files.

can please help me with this .

harshtech123:
@jdegoes  adding support for native and tests as well . i am on it !

harshtech123:
@jdegoes  can i only go for jvm support for now since , in scala -xml  the SAX APIs (for example, classes like org.xml.sax.EntityResolver and org.xml.sax.ErrorHandler), which are available on the JVM but not on Scala.js or Scala Native. In other words, even though Scala XML 2.3.0 is published for Scala Native (and Scala.js), certain XML parsing functionality (typically those based on the SAX parser) isn’t natively supported on those platforms.

jdegoes:
@harshtech123 This still needs a full and exhaustive test suite before consideration. Thank you!

harshtech123:
@jdegoes  i am on it , we will implement this for jvm only and have fully end to end tests for assertions, thank you

harshtech123:
@jdegoes full and exhaustive test suits are implemented now for each encoding , decoding and roundtrip , thank you !

jdegoes:
It is not passing Lint nor the build. Please take a look and fix!

harshtech123:
fixed lint and build issues , thank you !

harshtech123:
@jdegoes  could you please see my dm on discord , thank you !

harshtech123:
@plokhotnyuk please can we work on this ?

harshtech123:
@jdegoes could you please find some time to review this !

PR #811, opened 2025-02-28
Add test about union types decoding

guizmaii:


plokhotnyuk:
Thanks for the heads up about missing `&&`!

But decoding of union types works fine and reports error for invalid JSON.

PR #818, opened 2025-03-11
Add config to zio-schema-bson

WesselVS:
zio-bson-magnolia has support for a configuration (in addition to annotations). We need this support in zio-schema-bson as well. Right now, I've only implemented discriminator and class name transformer support, since this is what my company needs.

### DIFF
--- a/zio-schema-bson/src/main/scala/zio/schema/codec/BsonSchemaCodec.scala
+++ b/zio-schema-bson/src/main/scala/zio/schema/codec/BsonSchemaCodec.scala
@@ -114,7 +114,7 @@ object BsonSchemaCodec {
     def withClassNameMapping(classNameMapping: TermMapping): Config =
       copy(classNameMapping = classNameMapping)
 
-    def copy(
+    private[this] def copy(
       sumTypeHandling: SumTypeHandling = sumTypeHandling,
       classNameMapping: TermMapping = classNameMapping
     ): Config =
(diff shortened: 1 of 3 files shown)


plokhotnyuk:
Hi @WesselVS,

Thank you for your contribution to the project! I appreciate the effort you've put into this pull request. I have a few suggestions to help ensure our codebase remains maintainable and evolves smoothly over time.

When dealing with configuration that may change or expand, it's generally a good idea to avoid using case classes. This is because adding new flags or options can break binary compatibility of the auto-generated methods, which would require manual fixes and potentially introduce a lot of boilerplate code, as seen [here](https://github.com/zio/zio-json/blob/9f2437217b3d61bac4843705f0a35a9278014ba1/zio-json/shared/src/main/scala-3/zio/json/JsonCodecConfiguration.scala#L39-L156).

Instead, I recommend using a regular class with a private constructor and a builder API. This approach provides more flexibility and better supports future changes. You can find a great example of this pattern [here](https://github.com/plokhotnyuk/jsoniter-scala/blob/35d7720d19fc2b3541dd8bdd68de80c0d739b121/jsoniter-scala-macros/shared/src/main/scala-2/com/github/plokhotnyuk/jsoniter_scala/macros/JsonCodecMaker.scala#L23-L306).

If you're interested in learning more about designing libraries for easier evolution while maintaining backward binary compatibility, I highly recommend checking out [this insightful talk](https://www.youtube.com/watch?v=2wkEX6MCxJs).

Thank you again for your work, and I look forward to seeing your updates!

plokhotnyuk (approved):

plokhotnyuk:
@WesselVS I'm happy that it was helpful!

Please peek the latest release that is already available on the Maven Central and enjoy your work!

PR #823, opened 2025-03-30
Avro Codec decoding for GenericRecord Single field is now supported

### LINKED ISSUE #805: Avro GenericRecord with Single Field cannot be decoded. Decoder tries to decode as Primitive Value
I have a generic Kafka-Comsumer which works with Avro-Records. 

It is working fine but know I have a GenericRecord used as a Kafka-Key which cannot be decoded. 

The Avro Record contains a Single field "deviceId" of Type String 
`case class ViviRequestKey(deviceId: String)`

It is the first Occurrence where I have a Record with a single field, and that seems to be the Problem.

I decode it like this:
`AvroCodec.schemaBasedBinaryCodec[A].decodeGenericRecord(record)`

It throws the Exception:
`zio.schema.codec.DecodeError$MalformedFieldWithPath: class org.apache.avro.generic.GenericData$Record cannot be cast to class org.apache.avro.util.Utf8 (org.apache.avro.generic.GenericData$Record and org.apache.avro.util.Utf8 are in unnamed module of loader 'app')`

I followed the Code inside zio-schema-avro to see what is different compared to the other Records I decode.

The Working examples match on Schema.Record
- AvroCode.decodeGenericRecord:72
- AvroCode.decodeValue:76
- case record: Schema.Record[_] => decodeRecord(raw, record).map(_.asInstanceOf[A])  :202
- decodeRecord[A](value: A, schema: Schema.Record[_]):262`

With the "Problematic" Record (single field) it matches instead Schema.CaseClass1 (the case before) and ends up trying 
to decode it as a Primitive Value which fails and throws the Exception

- AvroCode.decodeGenericRecord:72
- AvroCode.decodeValue:76
- case s1 @ Schema.CaseClass1(_, _, _, _) => decodeCaseClass1(raw, s1) :201 
- decodeValue(raw, schema.field.schema).map(schema.defaultConstruct):234
- case lzy @ Schema.Lazy(_)              => decodeValue(raw, lzy.schema)
- case Schema.Primitive(standardType, _) => decodePrimitiveValues(raw, standardType):223

Having the "case record: Schema.Record[_]" before "case s1 @ Schema.CaseClass1(_, _, _, _)"
could be a solution but I am not sure if that's the right approach

harshtech123:
/closes #805 
/claim #805

updated logic to handle the record single field case properly also added test cases that asserts ! 
@jdegoes in test i define the case class as `case class ViviRequestKey(deviceId: String)` gives better understanding of how single field can be decoded , thank you so much !


ghostdogpr:
I am not involved with zio-schema, from what I can see the most active contributors are @plokhotnyuk and @987Nabil, maybe you guys can take a look?

harshtech123:
@987Nabil could you please take a look .

plokhotnyuk (approved):
From my perspective, it looks good. 

However, I should mention that I don't have personal experience with Avro yet. 

I'm primarily relying on the valuable feedback from contributors who are using zio-schema-avro in their projects.

harshtech123:
@jdegoes / @987Nabil  could you please help with collecting reward! thank you.

PR #824, opened 2025-04-13
Optimize `BinaryCodec::streamDecoder` implementations

guizmaii:
Use the new `ZPipeline.mapChunksEither` and `ZPipeline.mapEitherChunked` methods from ZIO to avoid using ZIO with pure code


987Nabil:
@guizmaii I think you can do the same for JsonCodec

guizmaii:
@987Nabil Indeed! Thanks 🙂

plokhotnyuk (approved):

PR #830, opened 2025-04-18
Independent en-/decoding of empty collections & options for JSON (#829)

### LINKED ISSUE #829: Independent en-/decoding of empty collections and options (null) for JSON
Experience in zio-json/http have shown, that how to deal with empty collections/nulls needs to be independently configurable for JSON en-/decoding.

We want to modify the config of the json codec so that we have 4 settings

* collection encoding
* collections decoding
* null encoding
* null decoding

The default should be, to write fields that are empty/null, but be graceful if they are missing

The decoder ignores the current available settings. That is also not a good status quo

987Nabil:
fixes #829

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,7 +40,7 @@ jobs:
       fail-fast: false
       matrix:
         java: ['temurin:17', 'temurin:21']
-        scala: ['2.12.20', '2.13.16', '3.3.4']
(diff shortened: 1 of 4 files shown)


plokhotnyuk (commented):
Current solution introduces a lot of problems in backward binary compatibility.

I would deffer it for 2.x version of zio-json together with removing dependency on magnolia.

PR #992, opened 2026-01-28
feat: Implement XML codec

### LINKED ISSUE #754: XML Codec

shruti2522:
this PR adds an initial `zio-schema-xml` implementation:

- as mentioned by @AdrielC, XSD can be generated directly from Schema[A], and the generated XSD is isomorphic to the underlying data type.
- XML encoding follows ScalaXB’s behavior closely, so values serialized via ScalaXB and via `zio-schema-xml` are structurally isomorphic and round-trip to equivalent values.
- round-trip parity is treated as the primary correctness check (Schema → XSD → ScalaXB and Schema → XML)
- default encoding favors smaller, practical XML and verbose type wrappers are optional and configurable.
- encoding and decoding are derived purely from Schema[A], without reflection or value-based guessing

### working demo
`XMLDemo.scala`:
```scala
package zio.schema.codec

import zio._
import zio.schema._

object XmlDemo extends ZIOAppDefault {

  case class Address(street: String, city: String, zip: Option[String])
  object Address { implicit val schema: Schema[Address] = DeriveSchema.gen }

  case class User(name: String, active: Boolean, address: Address, roles: List[String])
  object User { implicit val schema: Schema[User] = DeriveSchema.gen }

  override def run =
    for {
      _ <- Console.printLine("XML Codec Demo")

      user = User("Alice", true, Address("123 Main St", "Wonderland", None), List("admin", "editor"))

      // encode to XML
      codec   = XmlCodec.encode(User.schema)
      encoded = codec.encode(user)
      xmlStr  = new String(encoded.toArray)
      _       <- Console.printLine(s"\nEncoded XML:\n$xmlStr")

      // decode back
      decoded <- ZIO.fromEither(codec.decode(encoded))
      _       <- Console.printLine(s"\nDecoded Object success: ${decoded == user}")
      _       <- Console.printLine(s"$decoded")

      // generate XSD
      xsd <- ZIO.fromEither(XmlSchemaCodec.encode(User.schema))
      _   <- Console.printLine(s"\nGenerated XSD:\n$xsd")

    } yield ()
}
```
output: 

https://github.com/user-attachments/assets/3349c31d-83b0-4869-bb58-efa856980e08

fix #754 
/claim #754

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -307,13 +307,14 @@ lazy val zioSchemaXml = project
   .settings(
     libraryDependencies ++= Seq(
       "org.scala-lang.modules" %% "scala-xml"                % scalaXmlVersion,
-      "org.scalaxb"            %% "scalaxb"                  % "1.11.1",
+      "org.scalaxb"            %% "scalaxb"                  % "1.12.4",
       "org.scala-lang.modules" %% "scala-parser-combinators" % "2.4.0",
       "javax.xml.bind"         % "jaxb-api"                  % "2.3.1"
     ),
     Compile / scalaxb / scalaxbXsdSource := new File(baseDirectory.value, "src/test/resources"),
-    Compile / scalaxb / scalaxbDispatchVersion := "1.1.3",
-    Compile / scalaxb / scalaxbPackageName := "generated"
+    Compile / scalaxb / scalaxbPackageName := "generated",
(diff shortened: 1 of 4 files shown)


shruti2522:
Hi @jdegoes @AdrielC, just a gentle follow-up on this PR. It’s ready for review whenever you get a chance. thanks!

PR #1025, opened 2026-03-05
feat: add XML Codec module (#754)

### LINKED ISSUE #754: XML Codec

Abu1982:
## XML Codec for ZIO Schema

Closes #754

### Summary

This PR adds a new `zio-schema-xml` module that provides XML encoding and decoding support for ZIO Schema, following the same patterns as existing codecs (JSON, MessagePack, Protobuf, etc.).

### Implementation

**Module structure:**
- `XmlCodec.scala` - Entry point providing `BinaryCodec[A]` instances via implicit `xmlCodec[A]`
- `XmlEncoder.scala` - Encodes ZIO Schema values to XML strings with proper escaping
- `XmlDecoder.scala` - Decodes XML strings back to ZIO Schema values using `scala-xml`
- `XmlCodecSpec.scala` - Comprehensive roundtrip tests

**Supported types:**
- All primitive types (String, Int, Long, Boolean, Double, Float, BigDecimal, BigInteger, UUID, Char, Binary/Base64)
- All temporal types (LocalDate, LocalTime, LocalDateTime, Instant, Duration, ZonedDateTime, OffsetDateTime, OffsetTime, Period, Year, YearMonth, MonthDay, DayOfWeek, Month, ZoneId, ZoneOffset)
- Currency type
- Case classes (CaseClass0 through CaseClass22 via generic Record API)
- Sealed traits / enums (Enum1 through EnumN)
- Collections (List, Set, Map, Sequence)
- Optional values
- Either and Fallback
- Tuples
- Nested/recursive types via Schema.Lazy
- GenericRecord
- DynamicValue
- Transform schemas

**XML format:**
- Records encode fields as child elements: `<root><name>Alice</name><age>30</age></root>`
- Sequences use `<item>` elements: `<root><item>1</item><item>2</item></root>`
- Maps use `<entry><key>k</key><value>v</value></entry>` structure
- Enums use a `type` attribute: `<root type="Red"><value/></root>`
- Optional None renders as self-closing: `<field/>`
- XML declaration included: `<?xml version="1.0" encoding="UTF-8"?>`
- Special characters properly escaped (`&amp;`, `&lt;`, `&gt;`, `&quot;`, `&apos;`)

**Dependencies:**
- `org.scala-lang.modules:scala-xml:2.3.0` (JVM only, like MsgPack/Thrift/Avro/Bson modules)

### Testing

Roundtrip tests covering primitives, case classes, nested structures, collections, optionals, either, enums, tuples, temporal types, and XML special character handling.

### DIFF
--- a/zio-schema-xml/src/test/scala/zio/schema/codec/XmlCodecSpec.scala
+++ b/zio-schema-xml/src/test/scala/zio/schema/codec/XmlCodecSpec.scala
@@ -1,8 +1,6 @@
 package zio.schema.codec
 
-import zio._
 import zio.schema._
-import zio.test.Assertion._
 import zio.test._
 
 object XmlCodecSpec extends ZIOSpecDefault {


hriszc:
Follow-up on the XML codec thread: the patch comment I posted earlier is still the intended delta from my side, focused on `Option` encoding/decoding stability, whitespace preservation, and wiring the XML module into the standard test path.\n\nThe PR itself is green. If maintainers are open to folding those fixes into this branch, I can keep iterating from that direction rather than opening a duplicate PR.

PR #1042, opened 2026-03-13
fix: implement Deterministic hashCode for DynamicValue

### LINKED ISSUE #655: Schema deterministic hashcode for DynamicValue
I'm creating a ddd store for Entity and ValueObject.

ValueObject will use it's hashcode as key in the store (and I'll manage collisions storing a list instead of a single value object)

My current approach use something like this:
```scala
trait ValueObject[R]:

  def schema: Schema[R]

  extension (valueObject: R)
    def isSameValueObjectAs(other: R): Boolean =
      schema.diff(valueObject, other).isIdentical

    def valueHashCode: Long =
      schema.toDynamic(valueObject).hashCode
```

This way I can do something like `store.getValueObject[T: ValueObject](valueObject: T).filter(found => found.isSameValueObjectAs(valueObject).runHead`

However in my tests I observe the hashCode generated using DynamicValue changes, making my code fail.

samuel-asleep:
`DynamicValue.hashCode` was non-deterministic, breaking use as HashMap/HashSet keys and across JVM runs in two distinct ways.

## Root Causes

- **`Primitive[A]`**: Case class `hashCode` included `standardType.##`, which calls `System.identityHashCode` on `StandardType` singletons (plain `implicit object`s, not `case object`s) — unstable across JVM runs.
- **`Dictionary`**: Auto-generated `hashCode` used `Chunk.hashCode` (order-sensitive), but entries originate from iterating a `Map` with non-deterministic order. Two logically-identical dictionaries could produce different hashes and fail `equals` **within the same run**.

## Fixes

- **`Primitive[A]`**: Override `hashCode` using `standardType.tag.##` (a stable `String`) combined with `value.##` via `MurmurHash3`:
  ```scala
  // Before: standardType.## → System.identityHashCode, changes between JVM runs
  // After: stable tag-based hash
  DynamicValue.Primitive(42, StandardType.IntType).hashCode ==
    DynamicValue.Primitive(42, StandardType.IntType).hashCode // always true
  ```

- **`Dictionary`**: Override both `hashCode` (`entries.toSet.hashCode()`) and `equals` (`entries.toSet == otherEntries.toSet`) to be order-independent and mutually consistent:
  ```scala
  val d1 = DynamicValue.Dictionary(Chunk((k1, v1), (k2, v2)))
  val d2 = DynamicValue.Dictionary(Chunk((k2, v2), (k1, v1)))
  d1 == d2          // true  (was false)
  d1.hashCode == d2.hashCode  // true  (was false)
  ```

## Tests

Four property-based tests added to `DynamicValueSpec`:
- `Primitive` hashCode stability across separate instances
- `Primitive` hashCode stability via `toDynamic` round-trips
- `Dictionary` hashCode and equality are order-independent
- General hashCode/equals contract across arbitrary generated `DynamicValue`s

fixes #655 
/claim #655

### DIFF
--- a/tests/shared/src/test/scala/zio/schema/DynamicValueSpec.scala
+++ b/tests/shared/src/test/scala/zio/schema/DynamicValueSpec.scala
@@ -86,6 +86,51 @@ object DynamicValueSpec extends ZIOSpecDefault {
           }
         }
       ),
+      suite("hashCode")(
+        test("identical Primitive DynamicValues have the same hashCode") {
+          check(SchemaGen.anyLeafAndValue) {
+            case (schema, a) =>
+              val dv1 = schema.toDynamic(a)
+              val dv2 = schema.toDynamic(a)
+              assertTrue(dv1.hashCode() == dv2.hashCode())
+          }
+        },
+        test("Primitive hashCode is stable across separate instances") {
+          check(DynamicValueGen.anyPrimitiveDynamicValue(StandardType.IntType)) { prim =>
+            val copy = DynamicValue.Primitive(prim.value, prim.standardType)
+            assertTrue(prim.hashCode() == copy.hashCode())
+          }
+        },
+        test("Dictionary hashCode is order-independent") {
+          check(
+            DynamicValueGen
+              .anyPrimitiveDynamicValue(StandardType.StringType)
+              .zip(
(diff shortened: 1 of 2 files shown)


samuel-asleep:
@jdegoes @987Nabil please review !

